PR #3, opened 2026-01-13
fix: context indicator now uses recipe's max_model_len

nexty5870:
## Summary
- Fixed context indicator in `/chat` always showing 200K tokens instead of the actual `max_model_len` from the recipe
- Root cause: Model ID mismatch between `/status` (returns `model_path`) and `/v1/models` (returns recipe `id`)
- Added `root` field to `OpenAIModelInfo` containing the `model_path` for proper matching

## Changes
- `controller/models.py`: Added `root` field to `OpenAIModelInfo`
- `controller/routes/models.py`: Return `recipe.model_path` as `root` in `/v1/models` responses
- `frontend/src/app/chat/page.tsx`: Updated model lookup to match on both `id` and `root`

## Test plan
- [x] Start a model with a recipe that has custom `max_model_len` (e.g., 16384)
- [x] Open `/chat` and verify context indicator shows correct max (16.4K instead of 200K)
- [x] Verify `/v1/models` API returns `root` field with model path

### DIFF
--- a/controller/models.py
+++ b/controller/models.py
@@ -157,6 +157,7 @@ class OpenAIModelInfo(BaseModel):
     object: str = "model"
     created: int
     owned_by: str = "vllm-studio"
+    root: Optional[str] = None  # model_path for frontend matching
     active: bool = False
     max_model_len: Optional[int] = None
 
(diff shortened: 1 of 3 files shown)


0xSero:
Closing as part of Pi-Mono migration + controller-owned agent loop consolidation. Superseded by mainline cleanup.

PR #62, opened 2026-04-20
fix download backpressure listener cleanup

### LINKED ISSUE #56: Potential MaxListenersExceededWarning in download-manager during backpressure handling
## Summary

  While testing large model downloads, I observed a MaxListenersExceededWarning originating from controller/src/services/download-manager.ts in the backpressure path.

  The warning appears around this block:

  - controller/src/services/download-manager.ts:357
  - controller/src/services/download-manager.ts:359

  ## Observed Behavior

  When writer.write() returns false, the code awaits a Promise that attaches:

  - writer.once("drain", ...)
  - writer.once("error", ...)

  Under repeated backpressure, this can accumulate error listeners and trigger:
  MaxListenersExceededWarning: Possible EventTarget memory leak detected...

  ## Why this matters

  This is not necessarily a fatal crash, but it can:

  - create noisy logs during large/slow downloads
  - indicate listener accumulation over long-running download sessions
  - make diagnosis harder when real stream errors happen

  ## Reproduction (approximate)

  1. Start controller.
  2. Trigger a large multi-part model download.
  3. Watch controller logs.
  4. In some runs, warning appears while writing .part files.

  ## Suggested direction

  Consider a backpressure wait pattern that avoids attaching a fresh error listener each loop iteration (or ensures cleanup symmetry for both listeners in all paths).

saminkhan1:
## Summary

- fix the download backpressure path so `drain` and `error` listeners are always cleaned up
- keep writer errors visible across repeated backpressure cycles
- add regression coverage for repeated drain and error cases
- remove stale controller barrel exports that were breaking `controller` typecheck

## Why

Large or slow downloads can hit backpressure repeatedly. The previous loop added a new `error` listener each time `writer.write()` returned `false`, and that listener could be left behind when `drain` fired first. Over time that produced listener growth and `MaxListenersExceededWarning` noise in the controller download path.

The barrel export cleanup is supporting maintenance only. It removes re-exports of files that no longer exist so the controller checks stay green.

## Validation

- `cd controller && bun run typecheck`
- `cd controller && bun test controller/src/modules/downloads/stream-backpressure.test.ts`
- `cd frontend && npx next build`
- local staging Docker rebuild returned `200` from `http://localhost:3000`
- `cd frontend && npm run desktop:dist`

Closes #56.


0xSero:
Closing as superseded by the current main branch and v1.20.3 release. Please open a fresh PR from latest main if any work still applies.

PR #71, opened 2026-05-02
cleanup: prune internal docs and dead surfaces

0xSero:
## Summary

This PR is now focused on cleanup and LOC reduction:

- Removes internal/generated docs and repo-local tooling that are not relevant to end users.
- Removes dead jobs UI/API/store surfaces and the legacy chat DB usage reader.
- Keeps generated Factory/wiki artifacts ignored going forward.
- Adds targeted controller security hardening for engine launch binaries: rejects shell/interpreter ExLLaMA commands and non-`llama-server` llama.cpp overrides.

## Validation

- `bun test src/modules/engines/layers/backend-builder.test.ts`
- `bun run typecheck` in `controller/`
- `bun test` in `controller/`
- Earlier cleanup validation also passed frontend lint/test/build.

## Notes

The open security PR ideas for `exllama_command` and `llama_bin` were valid, but their branches target stale lifecycle paths and include unrelated churn. This branch implements the fixes directly in the current `controller/src/modules/engines/layers/backend-builder.ts` path with regression coverage.


frontend/src/app/api/agent/directories/route.ts
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    return (await stat(candidate)).isDirectory();
+  } catch {
+    return false;
+  }
+}
+
+export async function GET(request: NextRequest) {
+  const requestedPath = request.nextUrl.searchParams.get("path")?.trim();
+  const directoryPath = path.resolve(requestedPath || os.homedir());
```
0xSero, line 58: [P1] Lock this directory browser down before exposing it

This endpoint lets any caller choose an arbitrary absolute `path` and then returns the server's home path plus child directory names. On the deployed app that becomes unauthenticated filesystem enumeration of the machine running vLLM Studio; at minimum this should be limited to the local/Electron surface or gated to explicitly allowed roots before returning paths.

frontend/src/components/projects-nav-section.tsx
```diff
@@ -201,6 +230,147 @@ function formatRelative(isoString: string): string {
...
+          <div className="rounded border border-(--err)/30 bg-(--err)/10 px-3 py-2 text-xs text-(--err)">
+            {browseError || error}
+          </div>
+        )}
+        <div className="flex justify-end gap-2 border-t border-(--border) pt-4">
+          <Button variant="ghost" onClick={onClose}>
+            Cancel
+          </Button>
+          <Button
+            onClick={() => currentPath && onSelect(currentPath)}
```
0xSero: [P2] Don't ignore a pasted path when selecting

The modal copy says users can paste an absolute path, but the primary action always submits `currentPath`; if the user edits `draftPath` and clicks "Select this folder" without pressing Go/Enter first, we add the previously loaded directory instead of the typed one. Either make Select validate/use `draftPath`, or disable it when the draft differs from `currentPath`.

frontend/src/app/agent/_components/agent-workspace.tsx
```diff
@@ -570,6 +573,7 @@ export function AgentWorkspace() {
       if (!target) return; // wait for projects to load
       if (selectedProjectId !== target.id) {
         selectProject(target);
+        if (newParam === "1" && !sessionParam) return;
```
0xSero: [P2] Mark cross-project new-session URLs as handled before returning

For `/agent?project=<other>&new=1`, this returns before `handledNavRef.current` is updated. After `selectProject` updates state, the same URL is processed again and the `new=1` block appends another fresh tab, so one navigation creates two empty tabs. Set the handled key before returning here, or have the second pass skip appending after the project reset.

frontend/src/lib/agent/sessions-store.ts
```diff
@@ -37,6 +47,22 @@ function piSessionsRoot(): string {
     : path.join(homedir(), ".pi", "agent", "sessions");
 }
 
+async function loadPiSessionManager(): Promise<PiSessionManager> {
+  const modulePath = path.join(
+    process.cwd(),
```
0xSero: [P1] Resolve the pi session manager through the package graph

Building this path from `process.cwd()/node_modules/...` works in dev, but the standalone server starts from `.next/standalone/frontend`, where this package path is not present/traced. Loading a saved session then throws instead of replaying events. Use a normal import/`require.resolve` so Next traces the dependency, or explicitly copy/include the package in the standalone bundle.

.githooks/pre-push
```diff
@@ -0,0 +1,19 @@
...
+url="$2"
+
+while read local_ref local_sha remote_ref remote_sha
+do
+  if [ "$local_sha" = "0000000000000000000000000000000000000000" ]; then
+    continue
+  fi
+
+  if [ "$remote_sha" = "0000000000000000000000000000000000000000" ]; then
+    range="$local_sha"
```
0xSero: [P1] Don't validate the entire history on first push

When `remote_sha` is all zeroes for a new branch, `range="$local_sha"` makes `git log` walk every ancestor of the branch. This repo already has older non-conventional subjects, so the hook rejects otherwise valid first pushes. Compare against the default-branch merge base or another already-pushed boundary instead of passing the bare SHA.

release.config.cjs
```diff
@@ -6,8 +6,43 @@
 module.exports = {
   branches: ["main"],
   plugins: [
-    "@semantic-release/commit-analyzer",
-    "@semantic-release/release-notes-generator",
+    [
+      "@semantic-release/commit-analyzer",
+      {
+        preset: "conventionalcommits",
```
0xSero, line 12: [P1] Install the conventionalcommits preset used here

The release workflow currently runs semantic-release with `@semantic-release/commit-analyzer`, `@semantic-release/release-notes-generator`, and `@semantic-release/github`, but not `conventional-changelog-conventionalcommits`. Setting `preset: "conventionalcommits"` makes these plugins load that preset package, so the release job will fail unless the workflow/package installation is updated or this uses an installed preset.

.factory/hooks/desktop-rebuild.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PROJECT_DIR="${FACTORY_PROJECT_DIR:-/Users/sero/projects/vllm-studio}"
```
0xSero: [P2] Avoid committing a user-specific project fallback

This hook falls back to `/Users/sero/projects/vllm-studio`, so anyone without `FACTORY_PROJECT_DIR` gets marker/build behavior pointed at another user's absolute path. Since the hook also rebuilds and replaces `/Applications/vLLM Studio.app`, fail fast when `FACTORY_PROJECT_DIR` is missing or derive the repo root relative to the hook instead of baking in a local machine path.

0xSero (commented):
Deep review pass: found a few correctness/security issues plus some complexity that should be tightened before merge.

PR #161, opened 2026-06-22
fix(controller): reject request-controlled runtime job commands

OnlyTerp:
WP-005 hardening: runtime job APIs no longer accept caller-controlled executable or argv fields.

Changes:
- POST /runtime/jobs rejects payloads containing command or args with 400.
- All runtime upgrade wrapper routes reject payloads containing command or args with 400.
- Removed request-controlled command/args plumbing from engine-jobs, runtime-upgrade, and vllm-runtime.
- Operator-configured env/default upgrade paths remain functional.
- Recipe launch-command behavior is unchanged.

Verification:
- cd controller && bun run typecheck: passed
- cd controller && bun run lint: passed
- Focused controller tests passed
- Broader controller integration has one known unrelated observability fixture failure (host power readout non-zero)

### DIFF
--- a/controller/src/modules/engines/routes.ts
+++ b/controller/src/modules/engines/routes.ts
@@ -1,7 +1,7 @@
 import type { RouteRegistrar } from "../../http/route-registrar";
 import { delay } from "../../core/async";
 import { HttpStatus, badRequest, notFound, serviceUnavailable } from "../../core/errors";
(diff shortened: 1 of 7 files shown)


OnlyTerp:
Closing — absorbed via b3f8637b: the request-controlled command/args rejection in `routes.ts` is byte-identical on main, and the engine-spec install refactor removed the remaining env-command paths.

PR #162, opened 2026-06-22
fix(frontend): enforce registered agent filesystem roots

OnlyTerp:
WP-006 hardening: agent filesystem list/read APIs now operate only inside server-known registered project roots.

Changes:
- Agent FS list/read resolves the caller-supplied cwd against registered project roots and rejects unregistered roots.
- Registered roots and requested targets are canonicalized with realpath before containment checks.
- Traversal attempts and symlink escapes outside the resolved root are rejected.
- Legitimate reads/lists inside registered projects continue to work.

Verification:
- npm --prefix frontend run typecheck: passed
- npm --prefix frontend run lint: passed
- Focused frontend/API tests: cd frontend && npx tsx --test ../tests/frontend/agent-fs-root-boundary.test.ts: passed

### DIFF
--- a/frontend/src/features/agent/fs-store.ts
+++ b/frontend/src/features/agent/fs-store.ts
@@ -1,6 +1,7 @@
 import { existsSync, promises as fs, readdirSync, realpathSync, statSync } from "node:fs";
 import path from "node:path";
 import type { FsEntry } from "@/features/agent/filesystem-types";
+import { listProjectsFromStore } from "./projects-store";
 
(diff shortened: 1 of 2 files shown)


OnlyTerp:
Closing — main took a deliberate different direction in c823445d (`resolveWorkspaceRoot` accepts unregistered safe cwds and greylists system roots). The structural pieces (realpath resolution, traversal/symlink tests) are on main, so nothing is lost.

PR #163, opened 2026-06-23
feat(proxy): normalize tool request order for stable prefix-cache bytes

OnlyTerp:
## Summary
Normalize the order of tool definitions (and strip the default `tool_choice: "auto"`) before a request is serialized to the local inference server, so semantically-identical tool sets produce byte-identical request prefixes.

## Why (local-model)
vLLM / SGLang / llama.cpp use automatic prefix caching (KV-block reuse for shared prompt prefixes). Stable request bytes → higher prefix-cache hit rate → faster TTFT and less GPU recompute on repeated / long-context turns. This is an inference-efficiency win, not an API-cost concern.

## Changes
- `controller/src/modules/proxy/content-normalizer.ts`: `normalizeToolRequest` canonicalizes tool ordering; legacy `functions` are converted then sorted; nonstandard fields preserved.
- Applied on both the proxy forwarding path (`openai-routes.ts`) and the tokenization path (`tokenization-routes.ts`) so both serialize post-normalization.
- Adds `content-normalizer.test.ts`.

## Verification
- `cd controller && bun test src/modules/proxy/content-normalizer.test.ts` → 4 pass, 0 fail
- `cd controller && bun run typecheck` → clean

## Non-goals
- No nested JSON-Schema canonicalization (deferred — the small normalizer is sufficient for tool-order stability).
- Request semantics unchanged; only ordering + the redundant default are normalized.

### DIFF
--- a/controller/src/modules/proxy/content-normalizer.test.ts
+++ b/controller/src/modules/proxy/content-normalizer.test.ts
@@ -1,0 +1,123 @@
+import { describe, expect, test } from "bun:test";
+
+import { normalizeToolRequest } from "./content-normalizer";
+
+type Tool = {
+  type: "function";
+  function: Record<string, unknown>;
+};
+
+describe("normalizeToolRequest", () => {
+  const makeShuffledTools = (): Tool[] => [
+    {
+      type: "function",
+      function: {
+        parameters: { type: "object", properties: {} },
+        name: "charlie",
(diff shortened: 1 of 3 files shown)


OnlyTerp:
Closing — absorbed byte-identical: `content-normalizer.ts`, its test, and the tokenization-route wiring all match main.

PR #164, opened 2026-06-23
refactor: dedupe recipe launch-command extra-args into shared serializer

OnlyTerp:
## Summary
Extract the byte-identical launch-command extra-args logic that was duplicated between the frontend recipe preview and the controller launch builder into a single `shared/command-builder.ts`. Behavior-preserving.

## Changes
- New `shared/command-builder.ts`: internal-key sets, JSON-string keys, and the extra-args serialization, parameterized per-side via an options object (false-boolean handling, shell quoting, JSON normalization).
- `controller/src/modules/engines/process/backend-builder.ts` and `frontend/src/features/recipes/recipe-command.ts` consume it. Divergent / security behavior is kept per-side (model-runtime-default heuristics, `VLLM_STUDIO_ALLOW_CUSTOM_LAUNCH_COMMAND` env-gating, binary path resolution, the `enable_expert_parallelism` false-exception).

## Verification
- Before/after equivalence check: emitted launch argv is **byte-identical** across representative recipes (string / number / array / object / JSON-string `speculative_config` / boolean values, internal-key filtering, the `enable_expert_parallelism` exception, `llama_bin` security rejection, empty strings).
- `cd controller && bun run typecheck` → clean; `bun test src/shared/command-builder.test.ts` → 24 pass; frontend `typecheck` + `build` → clean.

## Non-goals
- No dry-run endpoint; the preview stays offline/string-formatted exactly as before.

### DIFF
--- a/controller/src/modules/engines/process/backend-builder.ts
+++ b/controller/src/modules/engines/process/backend-builder.ts
@@ -3,6 +3,10 @@ import { dirname, join } from "node:path";
 import type { Recipe } from "../../models/types";
 import type { Config } from "../../../config/env";
 import { stripForeignFlagKeys } from "../../../../../shared/contracts/engine-args";
+import {
+  appendExtraArguments as appendSharedExtraArgs,
+  appendLlamacppExtraArguments as appendSharedLlamacppExtraArgs,
+} from "../../../../../shared/command-builder";
 import { resolveBinary } from "../../../core/command";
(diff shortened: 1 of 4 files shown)


OnlyTerp:
Closing — main kept serialization inline in `backend-builder.ts` and evolved it (vLLM allowlist filtering, docker wrapping), so the extracted-module approach no longer fits the architecture.

PR #165, opened 2026-06-23
feat(usage): per-model and per-day prefix-cache hit-rate observability

OnlyTerp:
## Summary
Surface prefix-cache hit-rate in the usage dashboard, per model and per day. The aggregate already records `cache_read`/`cache_write` tokens; this exposes them in the breakdowns and UI.

## Why (local-model)
Lets you see whether local prefix caching is actually helping — e.g. after the tool-order normalization PR — and how the hit rate trends over time. Inference-efficiency signal, not API cost.

## Changes
- `controller/src/stores/inference-request-store.ts`: add `cache_read`/`cache_write`/`cache_hit_rate` to the `by_model`, `daily`, and `daily_by_model` breakdowns; extract a shared `calcCacheHitRate` (also used by the global hit-rate).
- `controller/src/modules/system/usage/pi-sessions.ts`: per-model / per-day cache accumulation in the parallel Pi-sessions aggregator.
- `shared/contracts/usage.ts` + frontend normalizer: additive cache fields (default 0).
- UI: a sortable "Cache hit %" column in the per-model table, plus a per-day readout and an "avg cache hit" summary stat in the daily chart.

## Verification
- `cd controller && bun test src/stores/inference-request-store.test.ts` → pass; `bun run typecheck` → clean.
- frontend `typecheck` + `lint` + `build` → clean.

## Non-goals
- Additive only; no contract removals. Frontend normalizer defaults missing fields to 0 for older payloads.


OnlyTerp:
Rebuilt on current main: main already had the cache DB columns and totals hit-rate, so this now adds only the novel parts — per-model/per-day cache SQL aggregation, contract fields, the Cache hit % sortable column, and an avg-cache-hit stat in the daily chart (adapted to the new year/month bucket layout). Reused main's hit-rate formula via a shared helper for consistency. check:quality green on push.

0xSero:
This is a good addition — additive contract fields, defaults handled in the normalizer, and the UI matches the existing usage-table conventions. It conflicts with main though: `pi-sessions.ts` was substantially restructured in the recent perf sweep, so the per-model/per-day cache accumulation needs re-applying onto the new aggregator (the `model-performance-table-model.ts` conflict is trivial; `shared/contracts/usage.ts` rebases clean). One nit: consider rendering "—" instead of "0.0%" in the cache-hit column when a model has zero cache read+write, so no-data doesn't read as a measured miss rate. Please rebase and I'll take another pass.

0xSero:
Keeping open for rework. Rebuild cache observability on the current request store, usage contract, normalizer, and UI. Cover per-model and daily fields and render an em dash when the denominator is zero.

0xSero:

PR #167, opened 2026-06-23
feat(agent): capability registry + read-only /capabilities view

OnlyTerp:
## Summary
A capability registry that unifies the currently-scattered capability definitions (MCP catalogue, stored MCP servers, skills, prompt templates) into one read-only view, plus a `/capabilities` page. This is a **non-breaking parallel layer** — existing registration/consumption paths are untouched.

## Changes
- `frontend/src/features/agent/capabilities/`: pure read-only adapters mapping each source into a unified `CapabilityEntry`; a non-enforcing `detectCollisions`; and a `CapabilityRegistry` class that shadow-rejects id collisions (with explicit `{ override: true }` opt-in), records an audit log, and builds from sources via `fromSources(...)`.
- A read-only `/capabilities` dashboard view: capabilities grouped by provenance + a collisions section. Reads the static MCP catalogue and GETs plugins/skills/prompt-templates; degrades gracefully if a source is unavailable.

## Verification
- frontend `typecheck` + `lint` + `build` → clean; adapter + registry + view-model unit tests pass.

## Non-goals
- No enforcement and no ownership shift yet — collision detection is non-enforcing and existing consumers are not rewired. This is the read-only foundation for a future unified action/capability layer.

### DIFF
--- a/frontend/src/app/capabilities/page.tsx
+++ b/frontend/src/app/capabilities/page.tsx
@@ -1,0 +1,5 @@
+import { CapabilitiesPage } from "@/features/capabilities/capabilities-page";
+
+export default function CapabilitiesRoute() {
+  return <CapabilitiesPage />;
+}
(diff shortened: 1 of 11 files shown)


OnlyTerp:
Closing — the MCP surface this registry unified (catalogue/store/types) was removed in 7b0714d0, and skills + prompt templates are already unified by `catalogue-effects.ts`. The concept no longer applies.

PR #159, opened 2026-06-22
fix(frontend): remove duplicate exported AggregatedSession type alias

OnlyTerp:
Fixes #158

Removes the `export` keyword from the route-local `AggregatedSession` type alias in `frontend/src/app/api/agent/sessions/all/route.ts`. The canonical client contract remains exported from `frontend/src/features/agent/session-contracts.ts`.

## Verification
- `npm run check:contracts` ✅
- `npm --prefix frontend run typecheck` ✅
- `npm --prefix frontend run lint` ✅

## Notes
- No runtime behavior change.
- No response JSON shape change.
- No route behavior change.
- Does not touch PR #151/#152/#155/#157.

PR #169, opened 2026-07-02
fix(contracts): dedupe Environment types into a shared contract

OnlyTerp:
## Summary

`npm run check:contracts` currently fails on main:

```
Duplicate exported type/interface declarations found:
- Environment
- EnvironmentEngineId
  controller/src/modules/environments/types.ts
  frontend/src/lib/types.ts
```

Same playbook as #159: one declaration, re-exports everywhere else.

- New `shared/contracts/environments.ts` holds the single `Environment` + `EnvironmentEngineId` declarations (registered in `validate-shared-contracts.mjs` contract names + allowed files).
- `controller/src/modules/environments/types.ts` and `frontend/src/lib/types.ts` become re-exports.
- The controller-only `EnvironmentId` brand + `asEnvironmentId` are dropped — they were used in exactly one place (the serializer's return cast) and the wire shape is `id: string` on both sides anyway.

#### Test plan

- [x] `node scripts/validate-shared-contracts.mjs` passes
- [x] `bun run typecheck` (controller), `tsc --noEmit` + `typecheck:desktop` (frontend)
- [x] `environments-routes` integration tests 4/4
- [x] knip, jscpd, ui-structure, cycles, depcheck, production build (pre-push `check:quality`)


OnlyTerp:
Closing — you beat me to it: 6532ce7c + 4fba5843 landed the same consolidation into shared/contracts/environments (including dropping the id brand, 80fc4fa9) while this was in flight. Race conditions of parallel agent loops.

PR #168, opened 2026-07-02
feat(engines): add ExLlamaV3 backend via TabbyAPI

OnlyTerp:
## Summary

The repo description promises "Control panel for VLLM, Sglang, llama.cpp, exllamav3" — this delivers the missing fourth engine as a fifth `EngineSpec`.

- **Contract**: adds `"exllamav3"` to the `Backend`/`EngineBackend` unions; `selected_runtime_target_ids` now keys off `EngineBackend` instead of an inline union.
- **Engine spec** (`specs/exllamav3-spec.ts`): launches TabbyAPI via `python <tabby_api_dir>/main.py --host --port --model-name <basename(model_path)>`, reusing the existing (previously dead) `tabby_api_dir` config key. Health path `/health`. Python resolution follows the mlx/sglang pattern (env override → managed venv → python3). Install follows the llama.cpp pattern: env-configured command via `LOCAL_STUDIO_EXLLAMAV3_UPGRADE_CMD`, clear message when unset.
- **Runtime targets/info**: TabbyAPI directory shows up as a configured runtime target with an installed/missing probe; `SystemRuntimeInfo.backends.exllamav3` reports dir presence + python version. Docker image pattern registered for target discovery.
- **Frontend**: ExLlamaV3 appears in the recipe backend picker, engine capabilities (conservative: general/model/environment/command tabs only), engines settings, and the server logs backend list.
- **Deliberately excluded**: `/environments` (no official pinned Docker image) and managed-venv installs (TabbyAPI is a checkout, not a pip package).

#### Test plan

- [x] `bun test tests/controller/integration/exllamav3-spec.test.ts` — 6/6 (buildCommand shape, unconfigured-dir error, extra_args passthrough, detectInvocation, served-model-name extraction, install-unconfigured error)
- [x] Existing engine tests still green (engine-managed-runtime-command, process-utilities, vllm-docker-image — 9/9)
- [x] `bun run typecheck` (controller + cli), `tsc --noEmit` + `typecheck:desktop` (frontend)
- [x] knip, jscpd, ui-structure, cycles, depcheck, production build (pre-push `check:quality`)

PR #170, opened 2026-07-02
feat(proxy): serve the Anthropic Messages API

OnlyTerp:
## Summary

Makes Local Studio a native `ANTHROPIC_BASE_URL` target — the integration llama.cpp (Jan 2026), Ollama, and LM Studio all shipped this year because Claude Code and other Anthropic-wire-protocol agents became the default way people use local models for real work.

- **`POST /v1/messages`** — translates Anthropic requests to OpenAI internally (llama.cpp's approach) and forwards to the running backend, so it works for every engine. Handles: system (string or blocks), text/image(base64+url)/tool_use/tool_result content blocks, tools + tool_choice (auto/any/tool), stop_sequences, temperature/top_p.
- **Streaming** — full Anthropic SSE event sequence (`message_start` → `content_block_start/delta/stop` → `message_delta` → `message_stop`) translated from OpenAI chunks by a pure state machine, including `input_json_delta` for streamed tool calls and text→tool block transitions. 15s `ping` keepalive (same Cloudflare-prefill rationale as the OpenAI path).
- **`POST /v1/messages/count_tokens`** — flattens the request and tokenizes upstream, with a fallback chain (`/tokenize` → TabbyAPI's `/v1/token/encode`).
- **Observability for free**: requests through `/v1/messages` are recorded by the existing inference accounting — per-model requests, tokens, TTFT show up in the usage page. None of the other local Anthropic endpoints have this.
- Errors are Anthropic-shaped (`{type:"error",error:{...}}`), and the model-not-running guard behaves exactly like the OpenAI route (proxy never launches models).

## Live validation — RTX 5090, TabbyAPI/ExLlamaV3 backend (branch of #168)

```
$ ANTHROPIC_BASE_URL=http://127.0.0.1:8188 ANTHROPIC_AUTH_TOKEN=local-studio \
    claude -p "Reply with exactly: LOCAL STUDIO OK" --model llama32-1b-exl3
"LOCAL STUDIO OK"        # real Claude Code v2.1.179, end to end
```

- Non-streaming `/v1/messages`: correct message shape, `stop_reason` mapping
- Streaming: verified full event sequence; `message_delta` carried real usage (`{"input_tokens":39,"output_tokens":19}`)
- `count_tokens`: `{"input_tokens":27}` via the TabbyAPI fallback
- Usage page after the Claude Code turn: `llama32-1b-exl3 | requests: 3 | tokens: 27171 | avg TTFT 931ms`

## Test plan

- [x] 10 unit tests: request translation (system/blocks/tools/tool_choice/images/tool_result splitting), response translation (text + tool_use + stop reasons), stream state machine (text stream, text→tool transition, usage propagation), SSE line parsing, prompt flattening
- [x] `bun run typecheck`, `node scripts/validate-shared-contracts.mjs`
- [x] Full pre-push `check:quality` gate
- [x] Live Claude Code session against the 5090 (above)

Note: fresh-clone `npm ci` (without `--legacy-peer-deps`) currently fails on main — the frontend lockfile looks out of sync with the pinned deps; `npm install` resolves. Happy to PR the regenerated lockfile separately if wanted.


0xSero:
Solid translation layer — clean state machine, real tests, and the live Claude Code validation is convincing. One blocker: main inlined `services/inference-client` into `http/local-fetch.ts` (3c146e97), so this compiles against a stale import path — GitHub reports it mergeable but typecheck fails post-merge; please rebase and update the import. Two asks with the rebase: pass through `reasoning_content` as Anthropic `thinking` blocks (or explicitly note the drop) — local reasoning models otherwise lose thinking output that the OpenAI route preserves via `reasoning.ts` — and consider mirroring the OpenAI route's strict-model 404 for unmatched models instead of forwarding upstream. With those this is ready.

0xSero:
Keeping open for rework. Rebuild Anthropic Messages on current local-fetch, model canonicalization, reasoning, strict-model, non-running, and accounting paths with complete nonstreaming, SSE, tool, error, and count-token coverage.

0xSero:

Issue #171, opened 2026-07-02
Design: multi-model concurrency (named instances, TTL, VRAM-aware admission)


OnlyTerp:
## Problem

The controller assumes exactly one running backend on one `inference_port` — `setActiveRecipe` evicts before launching, and the proxy guards against "the" running model. That locks Local Studio out of every 2026 multi-model workflow:

- chat model + embedding model (RAG)
- big coder + small fast summarizer (agent workflows)
- target + draft model (speculative decoding)
- vision model resident alongside text chat

llama-swap exists solely to solve this and is used by thousands — but it has no lifecycle GUI, no health/log/usage surface. Local Studio already owns all of that per recipe, so it can solve this **better**, not just equally.

## Proposed direction (incremental, zero default-behavior change)

**Phase 1 — instances instead of "the process"**
- `ModelInstance` = recipe id + assigned port + pid + state; controller keeps an instance table
- Port range config (`LOCAL_STUDIO_INSTANCE_PORTS=8700-8720`); `inference_port` stays the primary for back-compat
- `POST /launch/:recipeId` gains `{mode: "exclusive" | "shared"}` — default `exclusive` = today's evict-first semantics, so nothing changes for existing users
- Proxy routes by the request's `model` field to the matching instance's port; unknown model → existing 503 (proxy still never launches)

**Phase 2 — TTL + admission**
- Per-recipe `idle_ttl_seconds` (0 = pinned); idle sweeper on the existing Effect timer patterns; crash-loop budget stays per-recipe
- VRAM-aware admission: record each recipe's observed VRAM footprint at ready-time; before launching a shared instance, check free VRAM and either reject with a typed reason or evict the LRU unpinned instance (policy: `reject | evict-lru`)

**Phase 3 — UI**
- Recipes page shows a running *list* with pin/TTL controls; usage page already keys by model

## Non-goals

Multi-GPU placement logic, cross-machine orchestration, auto-launch-on-request (explicit-launch philosophy stays).

## Questions

1. Admission policy default: `reject` (safe) or `evict-lru` (convenient)?
2. Does the primary `inference_port` stay special ("selected" model) or become just another instance?
3. Are Docker environments in scope for instance accounting in phase 1?

Happy to build this in phases with hardware validation on a 5090/32GB workstation (chat 27B + draft 1B + embedder resident simultaneously as the demo) — but wanted your take on the direction before touching the launch state machine.

0xSero:
[...]

PR #172, opened 2026-07-03
feat(engines): multi-model concurrency — instances, port pool, proxy routing (Phase 1)

OnlyTerp:
## Summary

Phase 1 of #171: let the controller run more than one model at once.

Today, `EngineCoordinator.setActiveRecipe` hardwires `config.inference_port` and the proxy 503s any request whose model isn't the single process on that port. This PR adds:

- **Configurable port pool** via `LOCAL_STUDIO_INSTANCE_PORTS` env var (supports ranges like `8700-8710`)
- **Instance registry** — in-memory tracking of shared model instances with dead-pid pruning and orphan reconciliation
- **Opt-in shared launch** — `POST /launch/:recipeId` accepts `{"mode":"shared"}` to spawn on a pool port (default `"exclusive"` is byte-for-byte identical)
- **Proxy routing** — `/v1/chat/completions` checks primary port first, then routes to the matching instance's port
- **Instance management** — `GET /instances`, `POST /instances/:recipeId/stop`
- **UI integration** — `GET /recipes` shows instance-backed recipes as `"running"`
- **Cancel safety** — `/launch/:recipeId/cancel` tracks launch mode; shared cancellation only aborts the controller (never calls `setActiveRecipe(null)` which would evict the primary model)

## Design decisions baked in

1. **Launches serialized under the existing `switchLock`** — one model loads at a time, protecting VRAM; running instances are unaffected
2. **Primary port stays "the selected model"** — system metrics, tokenization, audio, and `/v1/models` keep primary-port semantics (answers #171 question 2)
3. **Docker recipes work** — the kill path already handles containers via `stopDockerContainersForProcesses`; VRAM admission accounting is Phase 2
4. **No persistence** — registry reconciles from live process scans on startup (same philosophy as existing `findInferenceProcess` approach)
5. **Process table is source of truth** — crashed instances disappear on next access via dead-pid pruning

## Commits

| Commit | Description |
|---|---|
| `feat(engines)` | Instance registry, port pool, shared launch, proxy routing, wiring |
| `fix(routes)` | Mode-aware cancel: shared abort does not call `setActiveRecipe(null)` |
| `test(controller)` | 15 integration tests including shared-cancel regression test |

## Files changed

| File | Change |
|---|---|
| `controller/src/config/env.ts` | `instance_ports` config field, `parseInstancePorts` helper |
| `controller/src/modules/engines/process/instance-registry.ts` | **New** — `InstanceRegistry` with reserve/release/reconcile |
| `controller/src/modules/engines/process/process-manager.ts` | `launchModel(recipe, portOverride?)` |
| `controller/src/modules/engines/engine-coordinator.ts` | Port-threaded health probing, `launchSharedInstance` method |
| `controller/src/modules/engines/engine-service.ts` | `LaunchInstanceResult` type, `launchSharedInstance` interface |
| `controller/src/modules/engines/routes.ts` | Shared launch path, `GET /instances`, `POST /instances/:recipeId/stop`, recipe status, mode-aware cancel |
| `controller/src/modules/proxy/openai-routes.ts` | Target-port resolution replacing guard block |
| `controller/src/app-context.ts` | Wire `instanceRegistry` into `AppContext` |
| `.env.example` | Document `LOCAL_STUDIO_INSTANCE_PORTS` |
| `tests/controller/integration/fixtures.ts` | Add `LOCAL_STUDIO_INSTANCE_PORTS` to env cleanup |
| `tests/controller/integration/instance-registry.test.ts` | **New** — 15 tests covering registry, routes, proxy routing, cancel safety |

## Verification status

| Check | PR branch | Clean `upstream/main` (worktree) | Status |
|---|---|---|---|
| `tsc --noEmit` (controller typecheck) | ✅ pass | ✅ pass | ✅ no regression |
| `validate-barrel-dir-siblings.mjs` | ✅ pass | ✅ pass | ✅ no regression |
| `validate-shared-contracts.mjs` | ❌ exit 1 | ❌ exit 1 (identical output) | ⚠️ **Pre-existing** — verified on `git worktree` of `upstream/main` |
| 4 regression test suites (20 tests) | 19 pass / 1 fail | 4 pass / 1 fail (identical) | ⚠️ **Pre-existing** — `observability-contracts` expects `current_power_watts: 0` but WSL returns real GPU readings; verified identical on clean main worktree |
| 15 new integration tests | ✅ 15/15 pass (includes shared-cancel regression) | N/A | ✅ |
| Root `npm run check` | ❌ not runnable | ❌ not runnable | ⚠️ **Environment issue** — WSL `npm` cascades to Windows npm which rejects UNC paths; sub-checks verified above |
| Live RTX 5090 smoke test | not run | N/A | ❌ **Unmet** — requires model files on disk; will run before merge |

## Exclusive mode is zero-change

With `LOCAL_STUDIO_INSTANCE_PORTS` unset:
- `POST /launch/:recipeId` with no body behaves identically
- Proxy routing returns 503 for non-running models as before
- No new ports are opened


0xSero:
The direction (opt-in shared instances, port pool, proxy routing by model, primary port stays special) matches the #171 design, and the cancel-safety regression test is genuinely valuable. But main has structurally moved: the launch path gained a `GpuLeaseRegistry` threaded through `setActiveRecipe` and `launchModel(recipe, LaunchModelOptions)`, and `engine-service.ts` no longer exists. As written, `launchSharedInstance` copy-pastes the old launch sequence and — rebased naively — would launch shared instances **without acquiring GPU leases** while exclusive launches require them, which is exactly the VRAM hazard Phase 2 is meant to prevent.

Rather than rebasing this diff, I'd like a revised Phase 1 that: (a) routes shared launches through the lease registry; (b) reconciles the instance registry eagerly at startup instead of lazily inside `list()` (first call currently returns stale/empty and races the unawaited `void reconcile()`); (c) verifies a pool port is actually free before reserving it. Happy to keep the proxy-routing and cancel-mode pieces largely as-is — and note the lease registry on main already partially delivers Phase 2's VRAM-aware admission, so the #171 design doc could use a refresh against current main too.

0xSero:
Keeping open for rework. Rebuild Phase 1 on current ComputeService and persisted InstanceStore, preserve exclusive primary behavior, reconcile records before routing, and cover collisions, capacity, cancel, restart, and model-to-port routing.

0xSero:

PR #174, opened 2026-07-09
fix(controller): resolve binaries cross-platform via Bun.which

### LINKED ISSUE #173: [BUG] GPU detection fails on Windows even with nvidia-smi on PATH - resolveBinary is POSIX-only
## Bug Description

On Windows the controller boots and serves fine, but the boot summary reports `gpu_tool=none detected` / `nvidia-smi=not found` even when `nvidia-smi.exe` is on `PATH` (`C:\Windows\System32`, the standard driver install location). Root cause is `resolveBinary` in `controller/src/core/command.ts`, which is POSIX-only:

1. `PATH` is split on `":"` — Windows uses `";"`, so every entry like `C:\Windows\System32` is shredded at the drive-letter colon
2. Candidates never try `PATHEXT` extensions — the file on disk is `nvidia-smi.exe`, but only the literal name `nvidia-smi` is checked
3. Only `"/"` marks an explicit path, so a backslash `NVIDIA_SMI_PATH` override fails too

## To Reproduce

1. Windows machine with an NVIDIA GPU (`nvidia-smi` resolves in any shell)
2. `cd controller && bun install && bun src/main.ts`
3. Boot summary prints:

```
gpu_tool=none detected
No GPUs reported by any monitoring tool; attempted: nvidia-smi=not found amd-smi=not found rocm-smi=not found intel-sysfs=unavailable
```

## Expected Behavior

`gpu_tool=nvidia-smi` and GPU metrics, as on Linux.

## Environment

- **OS**: Windows 11
- **Local Studio Version**: 0.3.2 (controller)
- **Backend**: n/a (fails at GPU monitoring, before any backend)
- **Model**: n/a

## Additional Context

I understand Windows isn't a release target (CI is ubuntu-only, engines target Linux CUDA / Apple MLX). I'm not proposing Windows support — this is a correctness fix to binary resolution that is a behavioral no-op on Linux/macOS. Since the controller is a Bun app, the hand-rolled PATH walk can be replaced outright with the built-in cross-platform `Bun.which()`, which also deletes ~20 lines. I have the fix ready (plus unit tests for `resolveBinary`; all controller gates green) and will open a PR.

luminary19:
## Summary

`resolveBinary` hand-rolled a POSIX-only PATH walk (`":"` delimiter, no `PATHEXT`, `/`-only explicit-path detection), so on Windows GPU monitoring silently reported no tools even with `nvidia-smi` on PATH. This replaces the walk with Bun's built-in cross-platform `Bun.which()` — less code, same public contract — while keeping the existing search-path semantics (`LOCAL_STUDIO_RUNTIME_BIN`/`SNAP` runtime dir first, then `PATH`, then home-bin fallbacks). Adds unit tests for `resolveBinary`. Fixes #173.

## Validation

```bash
cd controller
bun run test:unit        # 35 pass, incl. 7 new resolveBinary tests
bun run typecheck
bun run lint
bun run check            # knip + jscpd + depcheck + standards, all clean
bun run test:integration # unchanged vs. base
```

Verified live on Windows 11 (RTX 4070): boot summary now reports `gpu_tool=nvidia-smi` with correct GPU metrics. Linux/macOS behavior is unchanged by construction — there the delimiter is `:` and no `PATHEXT` applies, so the search resolves identically. The new tests are platform-agnostic (they create real temp executables) and run under the existing ubuntu CI.

## UI changes

None.

## Risks / rollout notes

- `Bun.which` requires the executable bit on POSIX; the old code accepted any regular file. Stricter and correct — a non-executable match could never be spawned anyway.
- Explicit paths now also honor `PATHEXT` on Windows (e.g. `NVIDIA_SMI_PATH=C:\...\nvidia-smi` finds `nvidia-smi.exe`).
- No new dependencies; net reduction of the hand-rolled resolution code.

### DIFF
--- a/controller/src/core/command.test.ts
+++ b/controller/src/core/command.test.ts
@@ -1,4 +1,4 @@
-﻿import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { delimiter, join } from "node:path";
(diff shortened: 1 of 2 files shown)


luminary19:
This is a small fix that only targets the binary resolution layer to make it OS-agnostic. Full windows run still doesn't work, but I think this nvidia-smi detection fix should be a good step.

PR #179, opened 2026-07-12
refactor: share controller extra-arg serialization

### LINKED ISSUE #144: Share controller extra_args lookup and CLI flag serialization
### Why
Controller launch code has multiple implementations of the same `extra_args` lookup and argument-emission rules. This is exactly where bugs show up when a recipe metadata field or internal key accidentally becomes a llama.cpp/vLLM CLI flag.

### Evidence
- `controller/src/modules/engines/process/backend-builder.ts` exports `getExtraArgument`, `appendExtraArguments`, and `appendLlamacppArguments`.
- `controller/src/modules/engines/process/process-utilities.ts` has a private `readExtraArgument` with the same snake_case / kebab-case fallback semantics.
- `appendExtraArguments` and `appendLlamacppArguments` duplicate the flag loop with separate internal-key sets and value conversion behavior.

### Cleanup
- Use one exported snake/kebab extra-arg lookup helper everywhere.
- Factor shared CLI flag emission into one serializer with backend-specific internal-key sets / value rules as data.
- Keep backend-specific flags explicit; do not broaden what gets passed through.

### Acceptance
- `npm --prefix controller run typecheck` passes.
- `npm --prefix controller run lint` passes.
- Existing process utility integration tests still pass.
- Recipe metadata/status/env-only fields are not emitted as CLI flags.
- vLLM, SGLang, llama.cpp, and MLX command construction remains byte-for-byte equivalent except for intentional removal of duplicated code.

lubluniky:
Closes #144

## Summary
- reuse one snake/kebab lookup across command, environment, and GPU lease paths
- share internal-key filtering and duplicate-flag handling while keeping the generic and llama.cpp value policies separate
- cover both serialization policies with exact argv tests

## Testing
- `cd controller && bun run check`
- `cd controller && bun run typecheck`
- `cd controller && bun run lint`
- `cd controller && bun run test:unit` (129 passed)
- `cd controller && bun run test:integration` (113 passed)

### DIFF
--- a/controller/src/modules/engines/argument-utilities.ts
+++ b/controller/src/modules/engines/argument-utilities.ts
@@ -13,6 +13,19 @@ export const extractFlag = (args: string[], flag: string): string | undefined =>
   return undefined;
 };
 
+export const getExtraArgument = (
+  extraArguments: Record<string, unknown>,
+  key: string,
+): unknown => {
+  if (Object.prototype.hasOwnProperty.call(extraArguments, key)) return extraArguments[key];
+  const kebab = key.replace(/_/g, "-");
+  if (Object.prototype.hasOwnProperty.call(extraArguments, kebab)) return extraArguments[kebab];
+  const snake = key.replace(/-/g, "_");
+  return Object.prototype.hasOwnProperty.call(extraArguments, snake)
+    ? extraArguments[snake]
+    : undefined;
+};
+
 const executableName = (value: string | undefined): string => {
   if (!value) return "";
   return value.split(/[\\/]/).filter(Boolean).at(-1)?.toLowerCase() ?? value.toLowerCase();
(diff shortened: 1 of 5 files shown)


0xSero (approved):
Verified argv-equivalence against main by hand and by running the full controller integration suite on this branch (113/113 + typecheck clean). The kebab/snake lookup, the `false`-emits-flag quirk with the expert-parallelism exception, and the llamacpp array-repeat/no-key-normalization differences all survive, and the new exact-argv tests pin them. One footnote: the `||` → `??` switch for `env_vars` lookup means a falsy non-nullish `env_vars` now shadows `env-vars`/`envVars` instead of falling through — inert in practice since the value must be an object, and the new behavior is the saner one.

PR #183, opened 2026-07-13
fix(build): normalize path separators in the shared-contracts check

### LINKED ISSUE #182: [BUG] check:contracts always fails on Windows - backslash relative paths never match the forward-slash allowlist
## Bug Description

`scripts/validate-shared-contracts.mjs` compares `relative(root, filePath)` against an `allowedFiles` set containing forward-slash entries (`"shared/contracts/recipes.ts"`, ...). On Windows `relative()` returns backslash paths (`shared\contracts\recipes.ts`), so no allowed file ever matches and the check reports every legitimate contract declaration as a violation.

## To Reproduce

1. Windows 11, clean tree at main
2. `npm run check:contracts`

```
Shared contract check failed. Move these declarations to shared/contracts:
- shared\contracts\controller-events.ts: ControllerEventType
- shared\contracts\observability.ts: GPU
- shared\contracts\recipes.ts: Backend
... (every contract file, ~60 lines)
```

3. Exit code 1 — `npm run check` can never pass on Windows.

## Expected Behavior

The check passes on a clean tree regardless of host path separator.

## Environment

- **OS**: Windows 11
- **Backend**: n/a

## Additional Context

One-line fix: normalize the relative path to forward slashes before the set lookup (also keeps the duplicate-declaration report consistent). Behavioral no-op on Linux/macOS — POSIX `relative()` output contains no backslashes to replace. Fix ready, will open a PR.

luminary19:
## Summary

`validate-shared-contracts.mjs` matched `relative()` output against a forward-slash allowlist; on Windows `relative()` yields backslash paths, so the check failed on every clean tree and `npm run check` could never pass there. The relative path is now normalized to `/` before comparison, which also keeps the duplicate-declaration report consistent across hosts. Fixes #182.

## Validation

```bash
npm run check:contracts   # passes on Windows 11; POSIX output has no backslashes to replace
```

## UI changes

None.

## Risks / rollout notes

None — `replaceAll("\\", "/")` is a no-op on POSIX `relative()` output.

### DIFF
--- a/scripts/validate-shared-contracts.mjs
+++ b/scripts/validate-shared-contracts.mjs
@@ -94,7 +94,7 @@ function walk(dir) {
 }
 
 function inspect(filePath) {
-  const rel = relative(root, filePath);
+  const rel = relative(root, filePath).replaceAll("\\", "/");
   const source = readFileSync(filePath, "utf8");
   collectExportedDeclarations(rel, source);
   for (const name of contractNames) {


0xSero (approved):
Confirmed the bug on main — `relative()` output at validate-shared-contracts.mjs:97 is compared against a forward-slash allowlist, so this can never pass on Windows. Normalizing at the single point `rel` is computed covers both the allowlist check and the duplicate-declaration report, and is a no-op on POSIX. The failing Dependency Review check is repo config (dependency graph disabled), not this PR.

PR #185, opened 2026-07-13
fix(controller): parse the CUDA UMD version banner from newer drivers

### LINKED ISSUE #184: [BUG] CUDA version undetected with driver 610+ unless nvcc is installed - nvidia-smi banner now prints CUDA UMD Version
## Bug Description

`controller/src/modules/engines/runtimes/runtime-info.ts` derives the CUDA version from the `nvidia-smi` banner line matching `CUDA Version: 12.8`. NVIDIA driver 610+ renamed the field: the banner now prints `CUDA UMD Version: 13.0` (UMD = user-mode driver). The regex no longer matches, so on driver 610+ machines without the CUDA toolkit installed (no `nvcc` to fall back to) the runtime report shows no CUDA version and the compatibility checks that key off it degrade. This is purely a driver-version string change — it applies to any platform and distro running the new driver series.

## To Reproduce

1. Any machine with NVIDIA driver >= 610 and no CUDA toolkit
2. `nvidia-smi | head -4` shows `CUDA UMD Version: 13.0` where older drivers printed `CUDA Version: ...`
3. `GET /runtime/info` → CUDA version `null` despite a healthy driver

## Expected Behavior

CUDA version detected from either banner spelling.

## Environment

- **OS**: Windows 11 (driver 610.47) — reproduces on Linux with the same driver series
- **Backend**: n/a (runtime info)

## Additional Context

Small parser fix: accept `CUDA (UMD )?Version:`; older drivers keep matching the old spelling. Extracting the parser into its own module also lets the llama.cpp installer reuse it later instead of duplicating the regex. Fix ready, will open a PR.

luminary19:
## Summary

NVIDIA driver 610+ renamed the `nvidia-smi` banner field from `CUDA Version:` to `CUDA UMD Version:`, so CUDA detection silently returned nothing on toolkit-less machines running the new driver series (the `nvcc` fallback masked it elsewhere). The parser moves to a small `cuda-version.ts` module and accepts both spellings; unit tests cover the old banner, the new banner, and no banner. Fixes #184.

## Validation

```bash
cd controller
bun run typecheck
bun run lint
bun test ./src/modules/engines/runtimes/cuda-version.test.ts   # 3 pass
```

Verified live on driver 610.47 (Windows 11, RTX 4070): `/runtime/info` reports the CUDA version again. Pre-610 banners are unaffected — the old spelling still matches.

## UI changes

None.

## Risks / rollout notes

None — pure parser widening; the extracted module is consumed only by runtime-info today.

### DIFF
--- a/controller/src/modules/engines/runtimes/cuda-version.test.ts
+++ b/controller/src/modules/engines/runtimes/cuda-version.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, test } from "bun:test";
+import { extractCudaVersion } from "./cuda-version";
+
+describe("extractCudaVersion", () => {
+  test("parses the classic nvidia-smi banner", () => {
+    expect(
+      extractCudaVersion(
+        "| NVIDIA-SMI 555.85    Driver Version: 555.85    CUDA Version: 12.5 |",
+      ),
+    ).toBe("12.5");
+  });
+
+  test("parses the driver 610+ UMD banner", () => {
+    expect(
+      extractCudaVersion(
+        "| NVIDIA-SMI 610.47    KMD Version: 610.47    CUDA UMD Version: 13.3 |",
+      ),
+    ).toBe("13.3");
+  });
+
+  test("returns null when no CUDA version is present", () => {
+    expect(extractCudaVersion("no gpu here")).toBeNull();
+  });
+});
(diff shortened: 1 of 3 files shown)


0xSero (approved):
Verified against main: the regex at runtime-info.ts:179 only matches `CUDA Version:`, so driver 610+ machines without nvcc silently report no CUDA — this widening fixes it while keeping the old banner. The extraction to cuda-version.ts is clean (both sync and async call sites use the import) and the three-case test suite runs under CI's `bun test src`.

PR #187, opened 2026-07-13
fix(controller): use per-user models dir and OS temp dir defaults on Windows

### LINKED ISSUE #186: [BUG] Controller defaults hardcode POSIX paths on Windows - /models and /tmp do not exist
## Bug Description

Two controller defaults assume a POSIX filesystem:

1. `controller/src/config/env.ts` defaults `LOCAL_STUDIO_MODELS_DIR` to `/models`. On Windows that resolves to `<current drive>:\models` — a directory that doesn't exist and that a normal user can't create without elevation. Every model listing and launch path starts from a dead directory.
2. `controller/src/core/log-files.ts` falls back to `/tmp` for launch logs when the data dir isn't writable — `/tmp` doesn't exist on Windows, so the fallback itself throws instead of rescuing the launch log.

## To Reproduce

1. Windows 11, `cd controller && bun install && bun src/main.ts`
2. `GET /models` → empty; the effective models dir is `\models` on the current drive
3. Make the data dir read-only to force the log fallback → ENOENT on `/tmp/...`

## Expected Behavior

A per-user models dir default (`~/models`) on Windows and the OS temp dir for fallback logs — the portable spellings of what the POSIX defaults intend.

## Environment

- **OS**: Windows 11
- **Backend**: n/a (fails before any backend)

luminary19:
## Summary

The models-dir default (`/models`) and the log fallback dir (`/tmp`) are POSIX literals; on Windows the former points at a non-existent drive root and the latter throws inside the fallback path itself. This defaults the models dir to `~/models` on win32 (POSIX keeps `/models`), replaces the `/tmp` literal with `os.tmpdir()` (which is `/tmp` on Linux), and comments the template default in `.env.example` so the platform default applies unless overridden. Fixes #186.

## Validation

```bash
cd controller
bun run typecheck   # clean
bun run lint        # clean
```

Verified live on Windows 11: boot summary shows `models_dir=C:\Users\<user>\models`, and model listing + llama.cpp launch work end-to-end. Linux behavior is unchanged by construction: the non-win32 models default stays `/models` and `os.tmpdir()` returns `/tmp`.

Note: the controller `test:unit` suite currently has POSIX-only fixtures that fail when run on Windows (unrelated to this change — those are addressed in a separate test-portability PR); this branch touches only `env.ts` and `log-files.ts` and does not alter any test.

## UI changes

None.

## Risks / rollout notes

- Anyone who copied the `.env.example` `LOCAL_STUDIO_MODELS_DIR=/models` line verbatim must uncomment it to keep an explicit override; the in-code POSIX default is identical, so nothing changes on Linux/macOS unless that override was being relied upon.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,7 +42,7 @@ LOCAL_STUDIO_INFERENCE_PORT=8000
 
 # Directory containing model weights. Must be writable: the controller tries to
 # create it on startup and warns if it cannot.
-LOCAL_STUDIO_MODELS_DIR=/models
+# LOCAL_STUDIO_MODELS_DIR=/models
 
 # Data directory for recipes and chat history
 LOCAL_STUDIO_DATA_DIR=./data
--- a/controller/src/config/env.ts
+++ b/controller/src/config/env.ts
@@ -1,7 +1,8 @@
 import { config as loadEnvironment } from "dotenv";
 import { Schema } from "effect";
 import { existsSync } from "node:fs";
-import { dirname, resolve } from "node:path";
+import { homedir } from "node:os";
+import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { loadPersistedConfig, type ProviderConfig } from "./persisted-config";
 import { parseBooleanFlag } from "../core/validation";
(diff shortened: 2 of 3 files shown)


0xSero (approved):
LGTM — the POSIX models-dir default stays literally `/models` by construction (the homelab relies on exactly that), and `os.tmpdir()` is the right call for the fallback log dir. One note for the record: on macOS `os.tmpdir()` returns $TMPDIR rather than /tmp, so the mac fallback log dir moves (self-consistent — old /tmp/vllm_*.log files just stop being listed); the "Linux behavior unchanged" claim is Linux-only, not all-POSIX. Dependency Review failure is our repo config, not this PR.

PR #189, opened 2026-07-13
fix(build): make repo bootstrap cross-platform (services link junction, node-script git hooks)

### LINKED ISSUE #188: [BUG] Fresh clone on Windows produces a broken dev environment - tracked symlink checks out as a plain file and git-hooks setup is POSIX-only
## Bug Description

Two independent POSIX assumptions break `git clone` → `npm install` on Windows:

1. `services/node_modules` is a **symlink tracked in git**. Windows git defaults to `core.symlinks=false`, so checkout materializes it as a small plain text file containing `../frontend/node_modules`. Module resolution under `services/agent-runtime` then fails (`effect`, `ws`, the pi SDK all unresolvable), and `frontend/scripts/link-services-node-modules.mjs` refuses to repair it — its guard sees "exists and is not a symlink" and exits. Even when the file is removed, `symlinkSync(..., "dir")` throws `EPERM` on stock Windows (directory symlinks need Developer Mode or elevation); a junction needs neither.
2. `setup:git-hooks` is `git config core.hooksPath .githooks && chmod +x .githooks/*`. `chmod` does not exist on Windows; the `|| true` in `frontend`'s `prepare` swallows the failure, so hooks are silently never installed.

## To Reproduce

1. Windows 11, stock git (`core.symlinks=false`), fresh clone
2. `npm --prefix frontend install`
3. `npm --prefix frontend run typecheck` fails with a wall of `error TS7006: Parameter '...' implicitly has an 'any' type` in `services/agent-runtime/src/*.ts` (dependencies unresolvable through the dead link)
4. `git config core.hooksPath` → unset; commits bypass `.githooks/*`

## Expected Behavior

`npm install` yields a working workspace and installed hooks, as on Linux/macOS.

## Environment

- **OS**: Windows 11
- **Backend**: n/a (fails before any backend)

luminary19:
## Summary

Fresh clones are broken on Windows in two ways: the tracked `services/node_modules` symlink checks out as a plain file under `core.symlinks=false` (and `symlinkSync(..., "dir")` needs privileges a junction doesn't), and `setup:git-hooks` shells out to `chmod`, silently skipped via `|| true`. This untracks the link (postinstall recreates it anyway), teaches `link-services-node-modules.mjs` to remove a stale checkout artifact and fall back to a junction on win32, replaces the `chmod` one-liner with a small node script (`scripts/setup-git-hooks.mjs`, invoked from frontend via `prepare-repo-hooks.mjs`), and self-heals the link before `dev`. Fixes #188.

## Validation

```bash
npm --prefix frontend install          # link created: junction on win32, symlink elsewhere
npm --prefix frontend run check:quality # lint + typecheck + desktop typecheck + cycles + ui-structure + knip + jscpd + depcheck + build — all green
node scripts/setup-git-hooks.mjs        # git config core.hooksPath -> .githooks
```

Verified live on Windows 11: fresh clone → install → typecheck green, hooks installed. Linux/macOS unchanged by construction — the symlink branch still runs first, and the node hook script performs the same two operations (`core.hooksPath` config + mark executable) the shell line did.

## UI changes

None.

## Risks / rollout notes

- `services/node_modules` leaves the git index; contributors with an existing checkout keep their working link (now gitignored). Postinstall recreates it on the next install.
- Hook files rely on git-preserved execute bits instead of `chmod +x` at install time — identical result on POSIX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ data/
 
 # Node.js
 node_modules/
+services/node_modules
 .next/
 .turbo/
 
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
     "format": "prettier --write \"src/**/*.{ts,tsx,js,jsx,css,md,json}\"",
     "format:check": "prettier --check \"src/**/*.{ts,tsx,js,jsx,css,md,json}\"",
     "depcheck": "depcheck --ignores=\"ws,hono,@hono/node-server,@types/ws,@types/react,@types/react-dom,@types/node,eslint-config-next,tailwindcss,@tailwindcss/postcss,depcheck,prettier,jscpd,knip,lint-staged,concurrently,electron-builder,madge\" --ignore-patterns=\"*.d.ts,.next/**,dist-desktop/**,desktop/dist/**\" --specials=next,webpack,babel --skip-missing",
-    "prepare": "cd .. && git rev-parse --git-dir >/dev/null 2>&1 && npm run setup:git-hooks || true",
+    "prepare": "node scripts/prepare-repo-hooks.mjs",
     "postinstall": "node scripts/patch-pi-ai-openai-text-boundaries.mjs && node scripts/link-services-node-modules.mjs",
     "check:deadcode": "knip",
     "check:dupes": "jscpd src",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,6 +3,7 @@
   "version": "0.2.9",
   "private": true,
   "scripts": {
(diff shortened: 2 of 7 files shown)


0xSero (approved):
LGTM — POSIX keeps the exact relative `dir` symlink (junction is win32-only), deploy-remote.sh already excludes and never depends on the tracked link, and the node hook script reproduces `core.hooksPath` + exec bits exactly (the 100755 modes cover fresh POSIX clones anyway). Untracking `services/node_modules` with postinstall + the `predev` self-heal is strictly more robust than shipping a symlink through git.

PR #193, opened 2026-07-13
fix(controller): derive model names and local-path detection platform-aware

### LINKED ISSUE #192: [BUG] Model names and local-path detection split on "/" — Windows paths misparsed everywhere they surface
## Bug Description

Several call sites derive a model's display name with
`model_path.split("/").pop()` and detect "is this a local file?" with
`startsWith("/")`. With `C:\Users\me\models\foo.gguf`:

- the "name" becomes the entire path (no `/` to split on), leaking full local
  paths into `/models`, metrics, and usage labels;
- local-path detection fails (`C:\...` doesn't start with `/`), so a local
  gguf is treated as a HuggingFace repo id and download/launch logic takes the
  wrong branch;
- llama.cpp runtime targets given a backslash binary path are classified as a
  `system` reference instead of a `binary` target.

## To Reproduce

1. Windows 11, create a llamacpp recipe with
   `model_path=C:/Users/<user>/models/<model>.gguf`
2. `GET /models` / usage views show the full path as the model name
3. Point a runtime target at `C:\...\llama-server.exe` → classified `system`,
   probe fails

## Expected Behavior

`path.basename()` semantics for names and `path.isAbsolute()` (plus drive/UNC
root checks) for local detection — identical results on POSIX.

## Environment

- **OS**: Windows 11
- **Backend**: llamacpp

luminary19:
## Summary

Model display names came from `split("/").pop()` and local-model detection
from `startsWith("/")` — both wrong for Windows paths (full path leaked as the
name; local ggufs treated as HF repo ids; backslash llama.cpp targets
misclassified as `system`). Adds `modelBasename` (`core/paths.ts`) used by
models/metrics/usage routes, switches detection to `isAbsolute` with platform
root checks, and classifies backslash binary paths as `binary` targets.
Fixes #192.

## Validation

```bash
cd controller
bun run typecheck && bun run lint && bun run check && bun run test:unit
bun run test:integration
```

New behavior-matrix test covers POSIX paths, Windows paths, and HF repo ids on
all platforms (repo ids keep their `org/name` split semantics). Verified live
on Windows 11: model list shows `<model>.gguf`, local gguf launches without a
HF lookup. POSIX outputs identical by construction.

## UI changes

None (names shown in existing views simply become correct on Windows).

## Risks / rollout notes

- HF repo ids (`org/name`) intentionally keep forward-slash semantics —
  covered by the test matrix.


0xSero:
Keeping open for rework because #192 still reproduces. Replace this historical branch with one portable path helper and table-driven POSIX, drive, UNC, slash-style Windows, and Hugging Face cases.

0xSero:

PR #195, opened 2026-07-13
fix(controller): reject malformed transcription uploads with a 400

### LINKED ISSUE #194: [BUG] Malformed multipart to /v1/audio/transcriptions returns a 500 instead of a 400
## Bug Description

`POST /v1/audio/transcriptions` parses the body with
`boundedFormData(...)` (`controller/src/modules/audio/routes.ts`), whose
inner `Request.formData()` rejects on malformed multipart (bad boundary,
truncated body, wrong content type). Nothing maps that rejection to a client
error, so the route 500s on caller mistakes — any OpenAI-compatible client
with a broken multipart encoder sees "server error" instead of "your request
is malformed". Applies to every platform.

## To Reproduce

```bash
curl -s -X POST http://127.0.0.1:8080/v1/audio/transcriptions \
  -H "Content-Type: multipart/form-data; boundary=nope" --data "not-multipart" \
  -w "\nHTTP:%{http_code}\n"
# HTTP:500
```

## Expected Behavior

`400` with the existing error envelope (`invalid_multipart`), consistent with
the route's other client-error responses (`file_missing`, `file_too_large`).

## Environment

- **OS**: any
- **Backend**: n/a (STT route)

luminary19:
## Summary

Malformed multipart on `/v1/audio/transcriptions` escaped
`boundedFormData()` as an unhandled rejection → 500. The parse now maps
non-size failures to `SttIntegrationError(400, "invalid_multipart")` while
rethrowing `RequestBodyTooLargeError` so the 413 path is unchanged. Fixes #194.

## Validation

`bun run typecheck` and `bun run lint` clean. Verified live against a running
controller:

```
# malformed multipart (bad boundary)
POST /v1/audio/transcriptions  ->  400 {"code":"invalid_multipart", ...}
# valid multipart, missing file field (unchanged)
POST /v1/audio/transcriptions  ->  400 {"code":"file_missing", ...}
```

The new contract assertion also lives in
`controller/tests/integration/system-studio-contracts.test.ts` and runs under
CI. Platform-agnostic change.

## UI changes

None.

## Risks / rollout notes

None — error mapping only; success and 413 paths untouched.


0xSero:
Closing as superseded. Malformed multipart mapping already landed independently and remains present on current dev and main. Modern regression coverage can be a small follow-up.

PR #197, opened 2026-07-13
fix(controller): size single-file gguf models in the vram calculator

### LINKED ISSUE #196: [BUG] VRAM calculator 404s on single-file GGUF models — only HF-style model directories are sized
## Bug Description

The VRAM calculator resolves a model to a directory and sums its weight
files. A llama.cpp recipe pointing at a single `.gguf` FILE (the primary
llama.cpp workflow) fails the directory lookup and the endpoint returns 404 ("Model weights not found"),
so the pre-launch VRAM estimate is unavailable exactly where it's most useful
(quantized single-file models).

## To Reproduce

1. Any platform, llamacpp recipe with `model_path=/models/foo-Q4_K_M.gguf`
2. VRAM calculator request for that recipe → 404 "Model weights not found"

## Expected Behavior

Single-file models are sized from the file itself (weights term; KV-cache
term stays unavailable without a config.json and is reported as such).

## Environment

- **OS**: any
- **Backend**: llamacpp

luminary19:
## Summary

The VRAM calculator only understood model directories; single-file GGUF
paths — the standard llama.cpp shape — 404'd ("Model weights not found") instead of being sized. File
paths are now sized directly for the weights term. Fixes #196.

## Validation

`bun run typecheck` and `bun run lint` clean. Verified live against a running
controller with a single-file gguf inside `models_dir`:

```
POST /vram-calculator {"model":".../ornith-1.0-9b-Q4_K_M.gguf","context_length":2048,"tp_size":1}
  ->  200 {"model_size_gb":5.24,"total_gb":7.77,"fits_in_vram":true,...}
```

Previously the same request 404'd ("Model weights not found"). A directory of
weight files (the existing HF path) is unchanged. The contract assertion also
lives in `controller/tests/integration/system-studio-contracts.test.ts` and
runs under CI. Platform-agnostic change.

## UI changes

None (the calculator UI simply starts returning results for gguf recipes).

## Risks / rollout notes

- KV-cache term is not estimated for ggufs (no config.json); the response
  marks it unavailable rather than guessing.


0xSero:
Closing as superseded. Single-file GGUF VRAM sizing already landed independently and is present on current branches.

PR #201, opened 2026-07-13
fix(agent): launch stdio MCP servers through cmd.exe and kill their child trees on Windows

### LINKED ISSUE #200: [BUG] stdio MCP servers fail to spawn on Windows and leak child trees on close
## Bug Description

1. stdio MCP server commands are spawned directly. On Windows the typical
   entries (`npx ...`, `uvx ...`) resolve to `.cmd` shims, which cannot be
   `spawn`ed without a shell (Node/Bun reject or ENOENT them). Every stdio
   MCP server configured with an npm-style launcher fails to start.
2. On session close only the direct child is killed; the actual server is a
   grandchild through the shim, so MCP server processes accumulate across
   sessions.

## To Reproduce

1. Windows 11, add any stdio MCP server (e.g. `npx -y
   @modelcontextprotocol/server-filesystem ...`) in agent settings
2. Server shows as failed/unavailable in `/agent`
3. With a direct-`node` workaround configured instead, close the session →
   the server process survives in Task Manager

## Expected Behavior

stdio MCP servers launch and are fully torn down on close, as on POSIX.

## Environment

- **OS**: Windows 11
- **Backend**: n/a (agent MCP)

luminary19:
## Summary

stdio MCP servers never started on Windows: their launchers (`npx.cmd` / `uvx.cmd`)
are `.cmd` shims that Node 24 refuses to spawn without a shell, and on close only
the direct child was signalled so the real server process leaked.

On win32 the client now:

- **Resolves the command first** (PATH + PATHEXT) and spawns `.exe`/`.com`
  targets **directly, with no shell** — only genuine `.cmd`/`.bat` shims are
  routed through `cmd.exe`.
- For those shims, builds a single command line under `cmd.exe /d /v:off /s /c`
  with proper **Windows CRT argument quoting plus double caret-escaping** and
  `windowsVerbatimArguments` (so Node does not re-quote). The double escaping is
  required because npm-style shims forward with `%*`, which triggers a **second**
  `cmd.exe` parse — a single-pass escaper is still injectable.
- Tears the process tree down with `taskkill /PID <pid> /T /F`.

Other platforms keep the direct spawn + SIGTERM path unchanged. Fixes #200

## Security

The first cut launched via `cmd.exe /c <command> <args…>` with unescaped args,
which a background review correctly flagged as command injection
(CVE-2024-27980 class): an MCP arg such as `a&calc.exe`, sourced from a
semi-trusted project `.mcp.json`, would break out. The hardened launcher above
neutralizes `& | < > ^ % !` across **both** cmd.exe parses. Verified live on
Node 24.16 by resolving the real `npx.CMD` and confirming benign args pass
through while `a&calc.exe`, `x|y`, `%PATH%`, `!v!` are all caret-escaped to
literals (e.g. `a&calc.exe` → `a^^^&calc.exe`), so `calc.exe` never executes as
a separate command. (Command *authorization* — a config naming `calc.exe` as the
command itself — remains a separate boundary, unchanged here.)

## Validation

`npm --prefix frontend run typecheck` clean (covers the agent-runtime sources);
`check:quality` (including the standalone build) runs as the pre-push gate. The
Windows launch behavior was verified live on Windows 11: a filesystem MCP server
over `npx` connects, its tools list and execute, and closing the session removes
the whole process tree.

## UI changes

None.

## Risks / rollout notes

- `.exe`/`.com` MCP commands now bypass the shell entirely (stricter, correct).
- Escaping is defensive; it does not authorize which commands may run — that is a
  separate concern for the MCP-config trust flow.


0xSero:
Keeping open for rework. Preserve the SDK protocol client, add secure global cmd handling and full Windows process-tree teardown, and prove metacharacter safety plus no surviving descendants.

0xSero:

PR #203, opened 2026-07-13
fix(agent): harden workspace-root validation for Windows paths

### LINKED ISSUE #202: [BUG] Agent workspace-root boundary checks miss Windows path realities — system dirs accepted, 8.3 short names falsely rejected
## Bug Description

`fs-store.ts` guards the agent's filesystem access by validating the
workspace root and checking every path stays inside it. Three Windows gaps:

1. The system-directory denylist covers POSIX roots (`/`, `/etc`, ...) but
   not `C:\`, `C:\Windows`, `C:\Program Files` — an agent workspace can be
   pointed at the OS itself.
2. UNC roots (`\\server\share`) pass validation but break every downstream
   relative-path computation.
3. Boundary comparison mixes `realpathSync.native` with plain
   `realpathSync`. The two disagree on 8.3 short names (`C:\Users\RUNNER~1`
   vs `C:\Users\runneradmin`), so legitimately registered projects under
   short-name paths (any CI temp dir; user dirs with spaces) false-fail
   "Path escapes project root".

## To Reproduce

1. Windows 11, `/agent`: set workspace root to `C:\Windows` → accepted
2. Register a project whose path goes through an 8.3 alias (e.g. point
   TMP/TEMP at `C:\Users\SHORTN~1\...`) → file operations fail with "Path
   escapes project root" despite being inside it

## Expected Behavior

System roots rejected on every platform; canonical-path comparison immune to
short-name aliasing.

## Environment

- **OS**: Windows 11
- **Backend**: n/a (agent fs)

luminary19:
## Summary

The agent fs boundary accepted `C:\`/`C:\Windows`-style roots and UNC shares,
and false-rejected paths that traverse 8.3 short-name aliases because the
boundary check mixed `realpathSync.native` with plain `realpathSync` (they
disagree exactly on short names). Extends the system-root denylist to Windows
roots and UNC, and canonicalizes all boundary resolutions with
`realpathSync.native`. Adds an e2e boundary test. Fixes #202.

## Validation

`npm --prefix frontend run typecheck` clean; `check:quality` runs as the
pre-push gate. The boundary regression suite
(`tests/frontend/regression/agent-fs-root-boundary.test.ts`) exercises the new
cases — the added "rejects windows system directories as cwd" and the
short-name/UNC handling pass. Reproduced the short-name false-failure by
pointing TMP/TEMP at an 8.3 alias; green after the fix. POSIX is unchanged:
`.native` is the same syscall path as `realpathSync`, and the denylist
additions can never match POSIX paths.

(One pre-existing case in that suite — "accepts access through a symlinked
project root" — fails on Windows only in its `rmSync` cleanup of a directory
symlink; it is not touched by this change and passes on Linux CI.)

## UI changes

None (invalid roots now produce the existing rejection message).

## Risks / rollout notes

- Anyone intentionally using a UNC workspace root loses it — it never worked
  downstream, it just failed later and less clearly.


0xSero:
Keeping open for rework. Update current fs-store and colocated tests with drive, UNC, native canonicalization, and system-root coverage, then validate in Windows CI.

0xSero:

PR #207, opened 2026-07-13
fix(agent): Windows-aware terminal shell, path recognition, file URLs, and session cwd encoding

### LINKED ISSUE #206: [BUG] Agent terminal and path/URL handling break on Windows — commands run through the wrong shell, drive letters mangled
## Bug Description

Four related path/shell assumptions in the agent surface:

1. Terminal commands are executed with a POSIX shell invocation; on Windows
   they land in cmd.exe with sh syntax, so most agent-issued commands fail.
   When Git Bash is installed (the common case for developers) it should be
   preferred, with an explicit override honored.
2. The embedded-browser input treats anything matching `^[a-z]+:` as a URL
   scheme — `C:\Users\me\index.html` parses as scheme `c` and navigation
   fails instead of opening the local file.
3. File URLs are built by string concatenation that strips the drive colon
   (`file:///C/Users/...`), which Chromium rejects.
4. Pi session discovery encodes the session cwd into an id; `C:` survives as
   a raw colon and collides with the id separator, so sessions on Windows
   never round-trip (agent history appears empty).

## To Reproduce

1. Windows 11, `/agent`: run `ls -la` in the terminal tool → cmd.exe error
2. Paste `C:\...\index.html` into the embedded browser → navigation error
3. Open a local file via the browser tool → `file:///C/...` rejected
4. Start a session, reload `/agent` → session list empty

## Expected Behavior

Terminal uses an appropriate shell, Windows paths are recognized as paths,
file URLs keep `C:`, and session ids round-trip.

## Environment

- **OS**: Windows 11
- **Backend**: n/a (agent)

luminary19:
## Summary

Agent terminal commands ran under the wrong shell on Windows, and three
path-shaped bugs mangled drive letters: embedded-browser input parsed
`C:\...` as a URL scheme, generated file URLs dropped the drive colon, and pi
session ids collided on `:` so session history never loaded. Terminal now
prefers Git Bash when present (override via settings honored), path detection
requires `[A-Za-z]:[\\/]` before URL parsing, file URLs keep the drive colon,
and session cwds are encoded before embedding. e2e tests cover cwd encoding
round-trips. Fixes #206.

## Validation

`npm --prefix frontend run typecheck` clean; the `pi-cwd-encoding` regression
test passes; `check:quality` runs as the pre-push gate. The Windows behavior was
verified live on Windows 11: the agent terminal runs POSIX-style commands via
Git Bash, local files open in the embedded browser, and sessions persist across
reloads. POSIX is unchanged: shell selection short-circuits to the existing
path, and the tightened scheme check cannot reject a real URL (schemes are ≥2
chars in the wild; `about:`, `data:`, `https:` all unaffected).

## UI changes

None.

## Risks / rollout notes

- Machines without Git Bash fall back to the previous behavior (cmd.exe) —
  degraded but unchanged from today.


0xSero:
Keeping open for rework. Reimplement the four Windows terminal, drive-path URL, and session-cwd fixes in a focused dev PR with current architecture and Windows acceptance.

0xSero:

PR #211, opened 2026-07-13
fix(controller): platform-aware process table, kill fallbacks, and install-job cancellation

### LINKED ISSUE #210: [BUG] Process discovery and kill paths are POSIX-only — `ps`, `sudo kill`, and `pgrep`-style probes do nothing on Windows
## Bug Description

The controller's process management shells out to POSIX tools:

1. the process inventory runs `ps -eo pid=,ppid=,stat=,args=` — no `ps` on
   Windows, so inference-process discovery (`findInferenceProcess`), stale
   process cleanup, and stop confirmation all see an empty world. A model
   launched by a previous controller run is invisible and can never be evicted.
2. Kill fallbacks run `sudo -n kill -SIGTERM <pid>` — no `sudo`/`kill`; a
   process that ignores `process.kill()` (or a permission failure) is
   unkillable, and install jobs leave orphaned child trees (pip/cmake keep
   running after cancel).
3. The running-process python probe reads a process's command line via `ps`
   (absent on Windows) and splits it on whitespace, so quoted
   `C:\Program Files\...` paths are mis-parsed and the backend is misidentified.

## To Reproduce

1. Windows 11, launch any llama.cpp/vLLM recipe, restart the controller
2. `GET /status` → `running: false` while `llama-server.exe` is still alive
3. `POST /evict` → reports success, process survives
4. Cancel a runtime install mid-way → `pip`/`cmake` children keep running

## Expected Behavior

Discovery, eviction, and install-cancel behave as on Linux.

## Environment

- **OS**: Windows 11
- **Backend**: llamacpp (any)

luminary19:
## Summary

Process discovery ran `ps`, kill fallbacks ran `sudo -n kill`, and install
cancel signalled only the direct child — all no-ops or leaks on Windows. The
process inventory now goes through a platform seam: `Get-CimInstance
Win32_Process` (UTF-8, JSON) on win32, the existing `ps` path elsewhere, with
command lines tokenized by the shared quote-aware splitter instead of split on
spaces. Kill fallbacks use `taskkill /PID <pid> /T /F` on win32 and skip the
meaningless `sudo` attempt; install-job cancellation kills the child tree; the
running-process python probe reads the command line via CIM on Windows. POSIX
paths are unchanged. Fixes #210.

## Validation

`bun run typecheck` and `bun run lint` clean; the new
`parseWindowsProcessInventory` unit test passes. Verified live on Windows 11
that `listProcessInventory()` returns the real CIM process table (377 processes,
correct pid/ppid, quoted command paths tokenized). The POSIX branch is
byte-identical to the previous `listProcessInventory` (Linux takes the unchanged
`ps` path), so CI is unaffected.

Note: the existing `process-manager-spawn` integration suite has Windows-only
fixture issues (extensionless fake binaries, a POSIX-shaped orphan-sweep
expectation) that are addressed by the separate test-portability PR; on Linux CI
that suite runs the unchanged POSIX path and passes.

## UI changes

None.

## Risks / rollout notes

- CIM via PowerShell adds ~100-300ms per cold snapshot on win32 only; the
  snapshot is cached for the poll tick to avoid repeated forks.
- `taskkill /F` is a hard kill — used only where SIGKILL was already the
  intent.


0xSero:
Keeping open for rework. Target the current compute launcher and supported native llama.cpp only: Windows ownership proof, process-tree stop and cancel, restart adoption, status, and evict with no orphaned child.

0xSero:

PR #213, opened 2026-07-13
fix(controller): install prebuilt CUDA llama.cpp on Windows with staged, digest-verified swaps

### LINKED ISSUE #212: [BUG] Managed llama.cpp install is build-from-source only — no Windows prebuilt path, and upgrades fail on a locked running binary
## Bug Description

The managed llama.cpp install path is build-from-source only, so it cannot
install llama.cpp on Windows:

1. `installManagedLlamacpp` unconditionally requires `git` and `cmake` on PATH
   and clones + configures + builds the repo. A stock Windows box has neither
   the toolchain nor the CUDA compiler cmake expects, so "Install llama.cpp"
   fails before it starts. There is no prebuilt-binary path.
2. Even when a binary is present, an in-place rebuild over a **running**
   `llama-server.exe` fails — Windows locks the executable of a live process,
   so an upgrade while a model is loaded aborts and can leave a half-written
   install.
3. `getConfigHelp` resolves the help binary from `config.llama_bin` /
   `llama-server` on PATH only. After a managed install (which drops the binary
   under the data dir, not on PATH) the engine's config-help panel stays empty
   even though a usable `llama-server.exe` exists.

## To Reproduce

1. Windows 11 with an NVIDIA GPU, no `git`/`cmake`/CUDA toolkit installed
2. Settings → llama.cpp → **Install** → fails with a missing-`git` error
3. Provide a prebuilt `llama-server.exe` out of band, load a model, then hit
   **Upgrade** → rebuild fails because the running binary is locked
4. With a managed binary installed but not on PATH, open the llama.cpp config
   help panel → empty / "Failed to fetch llama.cpp config"

## Expected Behavior

Installing/upgrading llama.cpp works on Windows without a build toolchain,
survives a locked running binary, and the config-help panel finds the managed
binary. Linux/macOS keep the existing build-from-source path unchanged.

## Environment

- **OS**: Windows 11
- **Backend**: llamacpp (CUDA)

luminary19:
## Summary

Adds a Windows prebuilt-CUDA install path for managed llama.cpp and makes
upgrades survive a locked running binary. On win32, `installManagedLlamacpp`
now queries the latest `ggml-org/llama.cpp` release, detects the driver's CUDA
version from the `nvidia-smi` banner, picks the newest compatible
`bin-win-cuda-*-x64` build plus its matching `cudart` runtime, downloads both,
verifies each asset's `sha256` digest, extracts with the bundled
`System32\tar.exe`, and installs by **staged rename swap**: extract to a staging
dir, rename the old install aside, rename staging into place, then delete the
retired copy. Because the swap never writes over the live executable, an upgrade
while a model is loaded no longer fails on a locked `llama-server.exe`; a still
-locked retired copy is detected up front with a clear "stop it and retry"
message. Downloads and extraction are bounded by timeouts.

The driver-CUDA banner parser is shared with the wave-1 `cuda-version.ts`
extractor (one regex, `CUDA (?:UMD )?Version`) instead of a second copy. Config
help now resolves through `resolveLlamaServerHelpBinary`, which falls back to the
managed install location when `llama_bin` is unset and the binary is not on
PATH, so the help panel works right after a managed install.

POSIX is untouched: non-win32 still requires `git`/`cmake` and builds from
source exactly as before. Stacked on wave-1 `cuda-umd-banner` (#185); rebased
onto `main`.

## Validation

`bun run typecheck` and `bun run lint` clean. The managed-llamacpp unit suite
(13 tests — asset selection, CUDA-version compatibility picking, digest gating,
help-binary resolution) passes.

Verified live on Windows 11 (RTX 4070): a real managed install downloaded
`llama-b9988-bin-win-cuda-13.3-x64.zip` + matching cudart, digest-verified,
extracted, staged-swapped, and the installed `llama-server.exe --version`
reported `version: 9988 (f4253ef96)`. With no `llama_bin` set,
`resolveLlamaServerHelpBinary` then resolved that managed binary and
`--help` exited 0. Total install ~156 s over the network.

The POSIX branch is unchanged, so Linux CI builds llama.cpp from source as
before.

## UI changes

None (the existing Install/Upgrade and config-help flows now succeed on
Windows).

## Risks / rollout notes

- Win32 install fetches from `api.github.com` and the release CDN; behind a
  proxy/offline box it fails with a clear error and the source-build /
  `LOCAL_STUDIO_LLAMA_BIN` / `LOCAL_STUDIO_LLAMACPP_UPGRADE_CMD` escape hatches
  still apply.
- Picks the newest CUDA build **at or below** the detected driver version; if
  the driver banner is unreadable it falls back to the lowest-CUDA build.
- The staged swap needs room for one extra copy of the install under the data
  dir during the swap; the retired copy is deleted immediately after.

Fixes #212.


0xSero:
Keeping open for rework. Add a digest-verified Windows llama.cpp prebuilt path, atomic replacement that handles a running executable, and managed config-help fallback while preserving POSIX source builds.

0xSero:

PR #217, opened 2026-07-13
fix(downloads): prevent backpressure listener leaks

### LINKED ISSUE #216: [BUG] Model download backpressure leaks writer error listeners
## Summary

The current model download backpressure loop leaves one `error` listener behind whenever `drain` wins. Repeated backpressure during a large or slow download therefore accumulates listeners and eventually emits `MaxListenersExceededWarning`.

This is the same failure mode originally reported in #56. The related PR #62 was closed as stale with a request to re-cut applicable work from current `main`; the affected loop is still present in `controller/src/modules/engines/downloads/download-manager.ts`.

## Reproduction

Running the current listener pattern through 12 drain cycles leaves 12 stale `error` listeners:

```text
{"drainListeners":0,"errorListeners":12}
MaxListenersExceededWarning: 11 error listeners added to [PassThrough]
```

## Expected behavior

- each backpressure wait removes both temporary listeners when either `drain` or `error` settles it;
- writer errors remain observable between drain cycles;
- repeated backpressure does not grow listener counts or emit a listener warning;
- focused regression coverage pins both repeated-drain cleanup and writer-error propagation.

lubluniky:
Closes #216

## Summary

- replace the per-drain listener pair with a wait that removes both temporary listeners when either `drain` or `error` settles it
- keep one lifecycle error tracker so a writer failure between backpressure cycles is surfaced instead of becoming unhandled or silently absorbed
- cover repeated drain cleanup, failed waits, and between-cycle writer errors

Before this change, 12 successful drain cycles left 12 stale `error` listeners and emitted `MaxListenersExceededWarning`. The regression test now pins one lifecycle listener during the transfer and zero listeners after disposal.

## Validation

```bash
cd controller && bun test src/modules/engines/downloads/stream-backpressure.test.ts tests/integration/download-manager-run.test.ts  # 7 passed
cd controller && bun run test:unit        # 145 passed
cd controller && bun run test:integration # 117 passed
cd controller && bun run typecheck
cd controller && bun run lint
cd controller && bun run check
npm run check
```

The pre-push frontend quality gate also passed, including the production Next.js build.

Environment note: a clean `npm ci` on current `main` hits the existing xterm beta peer-version conflict, so the root gate was run after `npm ci --legacy-peer-deps`; the lockfile is unchanged.

## UI changes

None. This only changes the controller's model-download stream lifecycle.

## Risks / rollout notes

Low. Download bytes, Range resume behavior, persistence, and progress updates are unchanged. The only behavioral difference is bounded listener lifetime and deterministic propagation of asynchronous writer failures. No migration or deployment step is required.


0xSero:
Closing as superseded. The equivalent backpressure fix is on both current branches; the focused test passed and repeated drain cycles left no temporary listeners.

PR #264, opened 2026-07-23
Add Atlas Cloud remote starter preset

binyangzhu000-sudo:
## Summary
- add Atlas Cloud as an OpenAI-compatible remote starter preset
- default the preset to https://api.atlascloud.ai/v1 and qwen/qwen3.5-flash
- add a focused controller test to keep the preset endpoint and model stable

## Tests
- bun test controller/tests/starter-presets.test.ts
- git diff --check
- git diff --cached --check
- Atlas live model catalog: 374 visible models, confirmed qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro

Note: attempted bun install --cwd controller --frozen-lockfile, but this local machine is out of disk space (ENOSPC while extracting tarballs), so full controller install/typecheck could not be run here.


0xSero:
Keeping open for rework. Atlas Cloud is aligned and the intended preset is small, but recreate it from current dev and verify the real keyed onboarding and first-chat surface; static config alone is insufficient.

0xSero:

PR #267, opened 2026-07-26
Compute layer: records-as-lease engines, benchmark-led onboarding, legacy launch core deleted

0xSero:
## What

Three arcs, one direction: replace the controller's tangled engine machinery with a small records-as-lease compute layer, make onboarding benchmark-led, and **delete the legacy core**.

### 1. Compute layer (`controller/src/modules/compute`, ~3.0k incl. bridge)
- **Records are the lease**: one JSON file per instance; `record.devices` IS the GPU hold; state derived (liveness → health → deadline); a 2s supervisor is the only reaper — dropping a record frees its devices by construction.
- Pure `plan()` per engine (vllm/sglang/llamacpp/mlx/exllamav3) with per-engine spelling tables; device selection translated in exactly one place (CUDA/ROCm/XPU env, docker flags).
- Launchers: process (group-reap, /proc start-token ownership) + docker (label-nonce ownership) behind one interface; capability-declaring device probes (NVIDIA/AMD/Apple/Intel + RAM/storage/thermals).
- One `LaunchFailure` union → `toHttp`/`toEvent` (no more substring-matched status codes).

### 2. Legacy core deleted (net −2,615 in controller/src)
`engine-coordinator`, `process-manager`, `process-utilities`, `process-inventory`, `backend-builder`, `launch-state` gone; `gpu-leases` 605 → 262 (store-backed shim, speech untouched); spec argument-builders gone. `compute/bridge.ts` answers the whole legacy surface — `/launch/:recipeId`, `/status`, proxy, metrics, audio, speech — from instance records. Preserved: custom `launch_command` (env-gated), MoE expert-parallel default, `LOCAL_STUDIO_READY_TIMEOUT_MS`, `llama_bin` validation, parser defaults, ps-scan runtime discovery.
**Bonus fixed by design**: models now survive controller restarts (records adopted from disk); ports are bind-probed so a squatting process can never be mistaken for a healthy launch.

### 3. Benchmark-led model picks + onboarding
- `shared/model-recommendations.json` (118 lines, 111 models) materialized from local.ai's measured runs: tested startup commands, sizes, pareto ranks, single-stream median speeds. Regenerate: `scripts/build-model-recommendations.mjs`.
- Dataset never ships to the client — `/api/setup/recommendations` prefilters to 8 display fields for the caller's rig.
- Onboarding: 6 wizard screens → 3 surfaces (Station / Model / Bring-up checklist); state machine untouched.

## Verification
- 78 controller tests + 43 frontend tests; tsc/eslint/knip/jscpd/standards + controller Effect-standards audit all green
- Live e2e on real hardware: legacy `/launch` → status → inference → duplicate-409 → `/evict` through compute with a real llama.cpp model; kill -9 → reaped ≤5s; restart-adoption; two instances sharing unified memory; MLX launch; dev Electron + browser walkthrough of the new onboarding
- Not yet exercised: docker launcher live (no Docker on the dev Mac), vLLM/SGLang real launches (validation rungs 1–3 against pop-os planned before deploy)


0xSero:
Superseded by #268 — this branch descends from the pre-split local lineage, so GitHub's diff dragged in 1,374 divergent-SHA commits whose content is already on main. #268 carries the same 5 new commits cleanly on top of current main.

PR #271, opened 2026-07-27
fix(ui): keep workspace notices clear of the composer

Dixith-dev:
This fixes the controller warning sitting behind the composer in Workbench.

The notice was absolutely positioned at the bottom-right with `z-30`, while the composer uses `z-[100]`. Since both were anchored to the same edge, the composer won the stacking order and hid most of the message.

I moved workspace notices to the top-right, just below the pane toolbar, and kept them above the workspace layers. That keeps the warning visible without covering the input. I also added a small regression test for the positioning.

Tested with:
- `npm run check`
- a packaged and installed dev desktop build
- the controller stopped temporarily to reproduce the warning and verify the final layout


0xSero:
Exact-head npm run check passed and the two-line layout direction is sound. Please rebase onto current dev, let CI run, and replace the source-string regex with a rendered or browser regression proving notices stay below the toolbar and do not overlap the composer.

Dixith-dev:
I’ve addressed this locally: the branch is rebased onto the current `dev`, and I replaced the source-string assertion with a Playwright regression that renders Workbench, forces the controller notice, and checks its real bounding box against the toolbar and composer.

Validation is green:
- `npm run check`
- `npm --prefix frontend run test:e2e:controller` (20 passed)

I’m blocked from updating the PR branch by the pre-push hook. Because this is a force-update after rebasing, the hook validates `old remote head..new head`, which includes 127 upstream `dev` commits. It rejects these two existing upstream commits for uppercase summaries:
- `80c8e04c feat(composer): ChatGPT-style goal mode`
- `f91beaa2 feat(panel): PR tab and Codex-style tab strip`

I haven’t bypassed the hook, so PR #271 is still on the old head. What’s the approved way to publish a rebased fork branch in this situation?

Dixith-dev:
Updated now: the branch is on the current `dev`, the browser layout regression is pushed at `31df4f46`, and the PR is ready for review. `npm run check` passes, and the controller Playwright suite passes all 20 tests. The new CI run is waiting for approval because it comes from a fork: https://github.com/sybil-solutions/local-studio/actions/runs/30954353423

0xSero:
Cherry-picked into #428 (minus the test file — tests are banned in this repo now). The bottom-3/right-3 z-30 layering fix applies as written.

PR #292, opened 2026-07-30
In-app updates from GitHub releases, file links that open anything, sidebar polish

0xSero:
## What's in here

**In-app updates (new)**
- Updater feed defaults to the public GitHub releases (`latest-mac.yml` + signed zip/dmg already ship there); electron-updater downloads in place, verifies the code signature, and `desktop:install-update` relaunches into the new version. User data is untouched.
- `useAppUpdate()` hook: installed version from the desktop bridge, newest release via a cached `/api/app-update` proxy of `Local-Studio-release.json` (no GitHub API rate limits).
- When a newer release exists: a download button appears in the sidebar footer (next to phone/settings), and **Settings → General** gains an *Application* block showing the installed version with an Update / Restart-to-update action. Web builds fall back to the DMG download.

**File links open anything on disk**
- Clicking an assistant file reference outside the session project used to silently do nothing. The Files panel now adopts that file's directory as a temporary root (with a back-to-project bar), expands `~`, follows `dir/` references, and surfaces read errors.
- Images render inline via a new `/api/agent/fs/raw` endpoint; PDFs/oversized binaries get Open (default app) / Reveal (Finder) actions via a home-confined `desktop:open-path` IPC.
- Fixes a pre-existing guard bypass: macOS symlinks `/etc`→`/private/etc`, which slipped past the literal system-root list and let fs routes serve `/etc/passwd`.

**Sidebar**
- Green "done" dot for runs that finished unviewed (distinct from blue unseen), compact relative ages on session rows, solid pin icons.

**Also on this branch**: release signing isolation CI (build → sign/notarize → publish), terminal scrollback replay, session rename, popup token tightening.

## Verification
- 109 bun tests pass (17 new: path resolution, workspace-root guard, version compare)
- lint / typecheck (web + desktop) / knip / madge / prettier clean
- `/api/agent/fs/raw` exercised against a real PDF (200, correct bytes) and against `/etc/passwd`, `/private/etc`, `../.ssh/id_rsa` (all rejected)


frontend/src/features/shell/use-app-update.ts
```diff
@@ -0,0 +1,120 @@
...
+      })
+      .catch(() => undefined);
+    syncDesktopPhase();
+    return () => {
+      cancelled = true;
+      if (pollTimer.current) clearTimeout(pollTimer.current);
+    };
+  }, [syncDesktopPhase]);
+
+  const updateAvailable = Boolean(
```
0xSero, line 94: **Dev false positive**: an unpackaged Electron run reports `app.getVersion()` from the repo's package.json (2.1.0), which is behind every published release — so dev builds would always show the update button and a forced check would error. `getRuntime` should expose `packaged` and this gate should require it.

frontend/src/features/shell/profile-footer.tsx
```diff
@@ -1,8 +1,38 @@
...
+      : update.phase === "working"
+        ? `Downloading v${update.latestVersion}…`
+        : `Update to v${update.latestVersion}`;
+  return (
+    <button
+      type="button"
+      onClick={update.startUpdate}
+      title={label}
+      aria-label={label}
+      className="flex h-7 w-7 shrink-0 items-center justify-center rounded-full bg-(--link) text-white transition-opacity hover:opacity-85"
```
0xSero: `text-white` is a hardcoded color in a tokenized UI — should be a token (e.g. the same fg used on primary buttons) so themes keep working.

frontend/src/features/settings/app-version-section.tsx
```diff
@@ -0,0 +1,49 @@
...
+    ? update.phase === "ready"
+      ? `v${update.latestVersion} is downloaded — restart to finish updating.`
+      : `v${update.latestVersion} is available on GitHub.`
+    : onLatest
+      ? "You are on the latest version."
+      : update.latestVersion
+        ? `Latest release: v${update.latestVersion}.`
+        : "Release check unavailable.";
+  return (
+    <SettingsGroup title="Application" description="Local Studio desktop app.">
```
0xSero: Description says "Local Studio desktop app." but this block also renders on the web build, where the row shows "Web UI". Wording should not assume desktop.

0xSero (commented):
Self-review pass — three real findings, will fix on this branch.

frontend/src/features/agent/ui/projects-nav/session-rows.tsx
```diff
@@ -122,12 +122,7 @@ export function ProjectRow({
         </button>
         <div className="absolute right-1.5 top-1/2 flex -translate-y-1/2 items-center gap-0.5">
           {onTogglePin ? (
-            <PinButton
-              pinned={pinned}
-              onToggle={onTogglePin}
-              target={project.name}
-              placement="inline"
-            />
+            <PinButton pinned={pinned} onToggle={onTogglePin} target={project.name} />
```
0xSero: **[Codex · medium] Finished runs lose their activity indicator**

`sessionActivity` now returns `finished`, but both row paths only pass `unseen` when the value is exactly `unseen`. When an unfocused run stops, the new `finished` state therefore renders no dot instead of the promised green done dot.

_(finding refers to `frontend/src/features/agent/ui/projects-nav/session-rows.tsx:422`)_

frontend/src/features/shell/use-app-update.ts
0xSero, line 94: **[Codex · medium] Web update fallback is unreachable**

The plain web app never supplies `currentVersion`, so `updateAvailable` is always false even when the release endpoint returns a newer version. Both update buttons are gated on this flag, making the documented DMG fallback in `startUpdate` impossible to invoke.

frontend/desktop/logic/update-manager.ts
```diff
@@ -30,25 +30,37 @@ function resolveFeedUrl(): string | null {
...
+      provider: "generic",
+      url: feedUrl,
+      channel: "stable",
+    });
+    return { ok: true, url: feedUrl };
   }
 
+  // Default feed: the public GitHub releases, which ship latest-mac.yml plus
+  // signed zip/dmg assets. electron-updater verifies the download's code
+  // signature against the running app before installing.
```
0xSero, line 47: **[Codex · medium] Dev builds are pointed at the stable release channel**

The default feed is hard-coded to the stable GitHub releases and `allowPrerelease` is later forced false without considering the packaged `localStudioChannel`. A signed Local Studio Dev build can therefore fetch and auto-install a stable Local Studio release instead of its documented dev prerelease.

_(finding refers to `frontend/desktop/logic/update-manager.ts:47`)_

0xSero (commented):
Review by OpenAI Codex CLI (codex exec, read-only sandbox) — findings posted verbatim as line comments (anchored to the nearest diff line where the original line was outside the diff).

frontend/desktop/logic/update-manager.ts
```diff
@@ -59,15 +71,7 @@ export async function checkForUpdates(force = false): Promise<DesktopUpdateSnaps
...
-  const feed = ensureFeedConfigured();
-  if (!feed.ok) {
-    const missingFeedState = {
-      status: "error",
-      message: feed.reason,
-    } satisfies DesktopUpdateSnapshot;
-    setUpdateState(missingFeedState);
-    return missingFeedState;
-  }
+  ensureFeedConfigured();
```
0xSero, line 87: **[Codex · low] Forced checks remain stuck in unpackaged builds**

The IPC caller passes `force=true`, bypassing this guard, but electron-updater still returns `null` for an unpackaged app without emitting a status event. `latestUpdateState` remains `checking`, so the renderer polls forever after clicking Update during desktop development.

_(finding refers to `frontend/desktop/logic/update-manager.ts:76`)_

frontend/src/features/agent/ui/projects-nav/session-nav-row.tsx
```diff
@@ -332,7 +333,7 @@ function SessionOptionsMenu({
 
   return (
     <div className={SESSION_MENU_CLASS} role="menu">
-      <MenuItem Icon={pref.pinned ? PinOff : Pin} onClick={run(onPin)}>
+      <MenuItem Icon={pref.pinned ? PinOffIcon : PinIcon} onClick={run(onPin)}>
```
0xSero, line 375: **[Codex · low] Session-age implementation is dead**

Rows explicitly render no age and never consume the newly added `timestamp`, `finished`, or `relativeAge` values; there is not even an age tooltip. If ages are intentionally omitted, the new props, import, and helper can be removed with identical functionality.

_(finding refers to `frontend/src/features/agent/ui/projects-nav/session-nav-row.tsx:303`)_

frontend/src/features/shell/use-app-update.ts
0xSero: Fixed in d74faef3 — `getRuntime` now reports `packaged` and the hook only trusts the version from packaged builds.

frontend/src/features/shell/profile-footer.tsx
0xSero: Fixed in d74faef3 — now `bg-(--color-primary) text-(--color-primary-foreground)`, same tokens as primary buttons.

frontend/src/features/settings/app-version-section.tsx
0xSero: Fixed in d74faef3 — description is now "Version and updates."

frontend/src/features/agent/ui/projects-nav/session-rows.tsx
0xSero: Fixed in d74faef3 — `finished` is threaded through SessionRow → SessionOpenTarget → SessionRowContent and renders a green `bg-(--ok)` dot (takes precedence over the blue unseen dot).

frontend/src/features/shell/use-app-update.ts
0xSero: Fixed in d74faef3 — the Settings row now offers a Download action whenever there is no installed-version signal (web build / dev run) and the release manifest is reachable, so the DMG path is exercised. The footer button stays desktop-only by design (web has no version to compare).

frontend/desktop/logic/update-manager.ts
0xSero: Fixed in d74faef3 — dev-channel builds (`isDevChannelBuild`) skip the stable GitHub feed in both `initializeAutoUpdates` and `checkForUpdates`; an explicit `LOCAL_STUDIO_UPDATE_URL` override still wins for feed testing.

frontend/desktop/logic/update-manager.ts
0xSero: Fixed in d74faef3 — a null result from `autoUpdater.checkForUpdates()` with state still `checking` now resolves to `idle` ("Updater unavailable in this build"), and unpackaged runs no longer surface the update UI at all.

frontend/src/features/agent/ui/projects-nav/session-nav-row.tsx
0xSero: Fixed in d74faef3 — dead `relativeAge` helper, its import, and the unused `timestamp` prop are removed; the `finished` prop it left behind is now actually rendered.

Issue #310, opened 2026-08-01
`ss` used in scripts/deploy-remote.sh is not available on macOS


amcfague:
## Summary

`scripts/deploy-remote.sh` relies on `ss` (iproute2) to check listening ports and resolve PIDs. `ss` is Linux-only — it is not available on macOS/OSX and cannot be installed there — so any of these code paths that execute on a developer's Mac will fail or silently misbehave.

## Occurrences

All in `scripts/deploy-remote.sh`:

- **Line 320** — `! ss -tlnp | grep -q ':8080\b'` — wait for the controller to stop.
- **Line 335** — `if ss -tlnp | grep -q ':8080\b'; then` — detect port 8080 still in use.
- **Line 337** — `ss -tlnp | grep ':8080\b' >&2 || true` — diagnostic dump of 8080 listeners.
- **Line 348** — `controller_pid=$(remote "ss -tlnp | sed -n '...pid=...' | head -1" ...)` — resolve controller PID.
- **Line 451** — `if ss -tlnp 2>/dev/null | grep -q ":${port}\b"; then` — port-based service status check.

## Notes

- These usages currently run inside `remote`/`<<'REMOTE'` heredoc blocks targeting the Linux host, where `ss` exists, so they work today. However, the pattern is fragile: any refactor that runs these checks locally (or reuse of these snippets in a local script) will break on macOS, and the dependency on `ss` is not obvious from the call sites.
- macOS provides no drop-in equivalent for `ss -tlnp`.

## Suggested fix

Use a portable alternative that works on both Linux and macOS, e.g. `lsof`:

```sh
lsof -nP -iTCP:8080 -sTCP:LISTEN
lsof -ti tcp:8080 -sTCP:LISTEN   # PIDs only
```

Alternatively, guard `ss` usage so it is only ever invoked on the remote host and never locally, and document that assumption.

0xSero:
Closing as completed because the affected deploy wrapper and all reported ss calls were retired from both current branches. PR #311 is obsolete and will close without merge.

PR #311, opened 2026-08-01
fix(deploy): support portable port readiness checks

amcfague:
## Summary

Port readiness, listener diagnostics, PID discovery, and stale-listener cleanup now work when Local Studio is invoked from or deployed to macOS. The same helpers run locally or are transmitted to remote Bash scripts, preferring Linux `ss` and falling back to macOS `lsof`.

Related: #310

## Validation

```bash
npm run check
bash -n scripts/deploy-remote.sh
shellcheck -e SC2029,SC2034,SC2086 scripts/deploy-remote.sh
git diff --check
```

Manually verified on macOS that the `lsof` path detects, identifies, and stops a real localhost listener through both local and transmitted remote-helper paths. A mocked `ss` path verifies Linux remains preferred when available.

## UI changes

None.

## Risks / rollout notes

macOS supplies `lsof`; supported Linux deployment hosts supply `ss`. This change makes port operations portable but does not alter the script's separate GPU or systemd-specific deployment behavior.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/scripts/deploy-remote.sh
+++ b/scripts/deploy-remote.sh
@@ -82,6 +82,15 @@ die() { fail "$@"; exit 1; }
 
 remote() { ssh $SSH_OPTS "$REMOTE" "$@"; }
 
+is_port_listening() {
+  local port="$1"
+  remote "if command -v ss >/dev/null 2>&1; then
+    ss -tlnp | grep -q ':${port}\\b'
+  else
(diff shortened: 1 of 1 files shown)


0xSero:
Closing as not aligned with current deployment architecture. The deploy wrapper this PR modifies was deliberately retired and is absent from dev and main.

PR #356, opened 2026-08-05
fix(desktop): keep updates inside Local Studio

0xSero:
## Summary
- promote the native-only update behavior after installing the repaired v2.9.8 baseline
- remove every renderer path that can open the browser for a desktop app update

## Validation
- PR #352 CI passed all required jobs against current `dev`
- `npm run check`
- baseline v2.9.8 installed with valid signature and bundled `app-update.yml`


0xSero:
Superseded by the clean current-main promotion PR; the dev-to-main comparison had a history-only conflict from the prior squash promotion.

PR #358, opened 2026-08-05
fix(desktop): preserve packaged Pi provider data

0xSero:
## Purpose

Prevent the packaged Local Studio Pi runtime from exiting on startup because its provider metadata was removed by the desktop resource filter.

## Changes

- narrow the standalone data exclusions to Local Studio project data only
- require a representative Pi provider data file during standalone and packaged-app validation

## Verification

- `npm run check`: frontend 123, controller 90, agent runtime 97; all passed
- `npm run test:integration`: 97 passed
- `npm --prefix frontend run desktop:pack`
- `npm --prefix frontend run desktop:smoke`
- deep code-sign verification
- packaged Pi 0.83.0 RPC `get_state`
- Litter phone-equivalent Local Studio E2E: 20 passed, 0 failed
  - start, streaming, reasoning, tools/filesystem
  - reconnect, two-way session visibility, compaction, and hydration

### DIFF
--- a/frontend/desktop/electron-builder.yml
+++ b/frontend/desktop/electron-builder.yml
@@ -62,7 +62,8 @@ extraResources:
     to: app/frontend/.next/standalone
     filter:
       - "**/*"
-      - "!**/data/**"
+      - "!data/**"
+      - "!frontend/data/**"
       - "!**/dist-desktop/**"
       - "!**/*.log"
   - from: .next/static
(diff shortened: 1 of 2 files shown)


0xSero (commented):
Self-review complete: no actionable findings. The resource filter is narrowly scoped to project data, the packaged Pi provider metadata is explicitly asserted, no application logic or unrelated scanner behavior changes, and the exact rebased commit passed the full CI matrix including desktop-package, frontend, controller, agent-runtime, dependency review, secret scan, and CodeQL.

PR #369, opened 2026-08-06
[Reliability] Make launch acceptance and cleanup attempt-atomic

### LINKED ISSUE #226: [Reliability] Make launch acceptance and cleanup atomic
## Audit metadata

- **Priority**: P1
- **Estimated effort**: M
- **Implementation risk**: MED
- **Category**: bug
- **Evidence baseline**: `07be5be7ce69bea0c3118744ab90d148b010fce0` (`origin/main` on 2026-07-17)

## Dependencies

None.

## Description and impact

Launch admission checks an idle snapshot, awaits process discovery, and only then marks launching. Concurrent requests can both be accepted. Same-recipe attempts can overwrite cancellation ownership, and an older `finally` can clear a newer launch's state. Replace check-then-set with a synchronous attempt lease whose token is required for cancellation and cleanup.

**Existing issue overlap**: closed issue #153 covers crash-loop budgeting, not concurrent admission or attempt ownership.

## Current state and reproduction

- `controller/src/modules/engines/lifecycle-routes.ts:17-35` reads idle and awaits `findInferenceProcess`.
- `lifecycle-routes.ts:48-51` creates the abort controller and marks launching after that await.
- `lifecycle-routes.ts:61-67` releases state by recipe ID rather than attempt identity.
- `controller/src/modules/engines/process/launch-state.ts:13-24` exposes unconditional `markLaunching`/`markIdle` and has no compare-and-set token.

To reproduce deterministically, delay `findInferenceProcess`, issue two POSTs to `/launch/<same-id>` before releasing the delay, then release both. Both requests pass the idle check. Different recipe IDs have the same admission race.

## Solution design

Create a synchronous launch lease API. `tryAcquire(recipeId)` returns an immutable `{ attemptId, recipeId }` only when idle; otherwise it returns the current snapshot/conflict. `release(attemptId)` succeeds only for the owner. Expose the active attempt for cancellation without leaking mutable state. Acquire immediately after validating that the recipe exists and before the first await; register the attempt's abort controller under `attemptId`; release it in `finally` by token.

Cancellation must abort the exact active attempt. It must not allow the old request or a stale cancel request to evict/clear a newer launch. If the current unscoped `setActiveRecipe(null)` is required for cleanup, keep the lease owned until that cleanup completes or move cleanup behind an attempt-aware coordinator operation. Use Effect primitives for any new async coordination, not timers or polling. Add no source comments.

## Verification commands

| Purpose | Command | Expected on success |
|---|---|---|
| Lease unit tests | `cd controller && bun test src/modules/engines/process/launch-state.test.ts` | exit 0 |
| Lifecycle integration | `cd controller && bun test tests/integration/runtime-recipe-contracts.test.ts` | exit 0 |
| Controller gate | `npm run check:controller` | exit 0 |
| Full integration | `npm run test:integration` | exit 0 |

## Scope

**In scope**

- `controller/src/modules/engines/process/launch-state.ts`
- `controller/src/modules/engines/process/launch-state.test.ts` (create)
- `controller/src/modules/engines/lifecycle-routes.ts`
- `controller/tests/integration/runtime-recipe-contracts.test.ts`

**Out of scope**

- Multi-model scheduling or launch queues; release 2.0 intentionally admits one launch.
- Crash-loop budgets, GPU lease policy, readiness timeouts, or engine command construction.
- Rewriting `EngineCoordinator` unless exact-attempt cancellation cannot be made safe otherwise; hit the STOP condition first.

### Step 1: Specify lease invariants in unit tests

Create tests proving only one synchronous acquire succeeds, conflicts preserve the active recipe, a wrong/stale token cannot release, the owner can release once, and a subsequent attempt gets a distinct token.

**Verify**: `cd controller && bun test src/modules/engines/process/launch-state.test.ts` -> tests fail until Step 2, then pass.

### Step 2: Implement token-owned launch state

Replace unconditional mutation methods with the lease API while preserving read-only `getState` and `getLaunchingRecipeId` consumers. Generate an opaque attempt ID with the repository's existing UUID facility.

**Verify**: `cd controller && bun test src/modules/engines/process/launch-state.test.ts` -> all pass.

### Step 3: Acquire before the first await and scope cleanup

In `registerLifecycleRoutes`, acquire after recipe lookup and before process discovery. Key abort/cancel ownership by attempt ID, reject conflicts with the existing 409 contract, and call token-aware release in every exit path. Ensure cancellation cannot act on a later attempt with the same recipe ID.

**Verify**: `npm run check:controller` -> exit 0.

### Step 4: Add deterministic concurrent route tests

Use a deferred promise/barrier in a test harness process-manager seam. Test same-recipe and different-recipe concurrent requests, stale cleanup after a newer acquisition, and cancel-versus-relaunch ordering. Do not use sleep-based race tests.

**Verify**: `cd controller && bun test tests/integration/runtime-recipe-contracts.test.ts` -> exactly one launch is admitted per barrier and all cases pass.

## Test plan

Unit-test the state machine separately, then integration-test HTTP status and service-call counts. Assert one accepted service invocation, one 409, state remains owned until completion, stale release is false/no-op, and cancellation targets the same attempt token.

## Acceptance criteria

- [ ] There is no await between successful admission and state reservation.
- [ ] Concurrent launch requests produce one accepted attempt and deterministic 409 responses.
- [ ] An old attempt cannot clear or cancel a newer attempt.
- [ ] Cancel addresses exactly the active attempt.
- [ ] Tests use barriers, not timing assumptions.
- [ ] `npm run check` and `npm run test:integration` exit 0.

## Risks and stop conditions

- Safe cancellation requires a public `EngineCoordinator` semantic change; report the required API and interleaving before expanding scope.
- The intended product behavior is queueing rather than 409; that is a separate scheduler design.
- Any path releases state by recipe ID instead of attempt token.
- Source drift or a verification command fails twice.

## Maintenance considerations

Attempt identity, not recipe identity, is the concurrency boundary. Review future retry, queue, or multi-model work for preservation of compare-and-set acquisition and token-owned cleanup.

---

Generated from Local Studio 2.0 main-branch audit proposal 007.

fettpl:
## Summary

- make launch admission and cleanup operate on an immutable attempt nonce
- serialize instance-record mutations across controller processes with a crash-recoverable SQLite transaction
- preserve atomic JSON replacement while making journal intent, replay, abort, and compaction ordering durable
- require caller-supplied attempt nonces on cancellation paths so delayed cancellation cannot target a replacement attempt
- keep the `speech` instance namespace reserved for its GPU lease

## Root cause

Launch ownership was scoped to a mutable instance name. Concurrent requests could pass preparation before a durable reservation existed, while cancellation and finalization could later act on whichever attempt occupied that name. File-based claims could also survive a controller crash and wedge future cleanup.

## Design

Admission synchronously persists a generated immutable attempt nonce before asynchronous discovery or preparation. Reserve, replace, release, cancellation, supervision, and failure cleanup compare against that nonce. Cross-process mutations acquire `BEGIN IMMEDIATE`, validate and replay any journal state, update the singleton mutex row, durably write the JSON record through an atomic replacement, and commit only after the file and directory operations succeed. Failed mutations write an explicit abort terminal before the transaction is released, so replay cannot resurrect an intent that did not complete.

Cancellation endpoints carry the attempt nonce; the bridge validates the recipe and never substitutes the current nonce. Stale cleanup can therefore stop only its own process reference and release only its own record. Legacy unhealthy records remain reclaimable, while generic compute launches cannot use the reserved `speech` name.

## Verification

- `npm run check` — passed, including repository and frontend static/quality gates and the production build
- manual concurrent launch and stale-cancellation probes — passed
- manual separate-process SQLite lock and abrupt-process recovery probe — passed; replace, release, and reacquire succeed after `SIGKILL`
- manual non-absence record-removal failure probe — passed; the error propagates and the transaction rolls back
- `git diff --check` — clean
- exact current-dev diff inspected against `dev` `fbc1278cbf13ab0ddd30836ae773380bdcf66701`
- no automated tests were added or run, per repository policy
- independent exact-head P0/P1 review — no findings

Closes #226

## Current-dev refresh (2026-08-20)

Published exact head `1dd840d6a1d6c6423578ee82e543d08d19e2d229`. The canonical `dev` commit `fbc1278cbf13ab0ddd30836ae773380bdcf66701` is the merge base and ancestor used for this refresh. The five changed files are limited to compute bridge, instance store, compute lifecycle/routes, and engine lifecycle routes; no frontend or packaged desktop behavior changed.


fettpl:
@0xSero, current-head refresh: [PR #369](https://github.com/sybil-solutions/local-studio/pull/369) is published at exact head `1dd840d6a1d6c6423578ee82e543d08d19e2d229` against canonical `dev` `fbc1278cbf13ab0ddd30836ae773380bdcf66701` (merge base and ancestor; clean diff). This is a production-only five-file repair for durable attempt ownership and serialized journal/SQLite mutations. `npm run check` passed; manual concurrent launch, stale-cancellation, SQLite lock/recovery, and rollback probes passed; no automated tests were added or run per repository policy; independent exact-head P0/P1 review found no findings.

0xSero:
Superseded: main's records-as-lease design (shipped and e2e-verified before this could land) already provides the acceptance atomicity this PR built with the nonce/journal/sqlite protocol, and merging both architectures produced a chimera. The one piece the lease design genuinely lacked — fsync before rename so power loss can't resurrect an empty lease file — was adopted into #428 with credit. Thanks for the durability catch.

PR #380, opened 2026-08-09
fix: surface provider models in Workbench

MarioMartinezII:
## Summary
- include enabled OpenAI-compatible provider catalogs in `GET /v1/models`
- preserve provider-qualified IDs such as `mai/gpt-5.6-sol` for existing chat routing
- share provider-catalog parsing between Studio settings and the OpenAI model catalog
- fail open when an enabled provider catalog is unavailable

## Test coverage
- enabled provider models appear with provider metadata
- disabled and unavailable providers are omitted without breaking the catalog
- 91 controller tests pass (0 failures)
- TypeScript, ESLint, Knip, duplication, dependency, and controller standards checks pass

## Deployment
- intended for the PHX Local Studio controller; no desktop bundle changes are required

### DIFF
--- a/controller/src/modules/models/routes.ts
+++ b/controller/src/modules/models/routes.ts
@@ -44,6 +44,7 @@ import { notFound } from "../../core/errors";
 import { findObservedInferenceProcess } from "../../core/function-observability";
 import { parseBooleanFlag } from "../../core/validation";
 import { fetchInference } from "../../http/local-fetch";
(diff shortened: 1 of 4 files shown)


0xSero:
Re-implemented in #428 together with #409 (the two branches collide in models/routes.ts): from this PR came the configurable discovery timeout and the richer provider-entry metadata (external flag, provider id, vision resolution). Fixes #410.

PR #395, opened 2026-08-11
feat(windows): add conservative Windows port

JoaoZaokk:
## Summary

- add a conservative Windows 11 x64 port while preserving the existing macOS and Linux implementations
- introduce platform-specific process, path, shell/PTY, GPU, controller lifecycle, and runtime handling
- add native Windows llama.cpp installation, controller startup integration, Electron packaging, and Windows CI
- expose vLLM and SGLang through an explicit WSL2 backend, including UI install, update, remove, launch, stop, status, and logs
- document the repository audit, independently testable port milestones, supported capabilities, and acceptance evidence

## Why

The desktop and controller contained POSIX and macOS assumptions around paths, shells, signals, process trees, services, runtime discovery, and packaging. Windows support needs platform adapters that retain current Darwin/Linux behavior rather than replacing those implementations.

vLLM and SGLang are not presented as native Windows engines. Their managed targets are explicitly labeled WSL2, while native Windows inference starts with llama.cpp and remote-controller behavior remains unchanged.

## Notable behavior

- Windows paths preserve drive letters, UNC paths, separators, and native shell behavior.
- PowerShell is preferred for the Windows PTY, with a cmd fallback.
- NVIDIA telemetry uses Windows-accessible tooling and has GPU-free parser tests.
- Multi-GPU Windows systems expose every detected adapter and the complete discrete VRAM pool.
- Hosted OpenAI-compatible providers are discoverable in Workbench; provider base URLs work with or without a trailing `/v1`.
- The DeepSeek starter uses the official public API instead of a private Tailscale endpoint.
- Local Windows service URLs advertise the configured loopback address rather than an unrelated machine hostname.
- Managed WSL2 engines stop only their own Linux process groups. They never call `wsl --shutdown` or terminate a distribution, so unrelated Docker/WSL workloads remain running.
- WSL2 runtime artifacts and caches are isolated per managed engine and removable from the UI.
- Existing macOS signing, notarization, packaging, and runtime behavior are left in place.

## Validation

- frontend: lint, typecheck, 141 unit tests, production Next.js build, and standalone verification
- controller: typecheck, lint, dependency/duplication/standards checks, and 123 unit tests
- agent runtime: build and 98 tests
- `git diff --check`
- live Windows provider smoke test:
  - the saved DeepSeek key listed `deepseek-v4-flash` and `deepseek-v4-pro`
  - a one-token OpenAI-compatible completion succeeded through the controller proxy
  - Workbench discovered both models through `/api/agent/models`
- live Windows multi-GPU smoke test:
  - RTX 3090 (24 GB) and RTX 3080 Ti (12 GB) were both reported
  - the setup pool resolved to 36 GB
- Windows acceptance on NVIDIA CUDA hardware:
  - vLLM 0.27.0 installed, served `HuggingFaceTB/SmolLM2-135M-Instruct`, returned an OpenAI-compatible chat completion, stopped, and was removed through the UI
  - SGLang 0.5.9 completed the same install/serve/chat/stop/remove flow
  - all Docker sentinel processes kept the same PIDs throughout engine stop and uninstall
  - temporary acceptance runtimes and isolated caches were removed afterward

No release was created and no package was published.


0xSero:
Consolidating the open Windows issues here so they land or fail together with this port — this draft is the only vehicle for them: #192 (model names / local-path detection split on "/", so Windows paths misparse), #210 (process discovery and kill paths are POSIX-only — ps/sudo kill/pgrep do nothing on Windows), #212 (managed llama.cpp install is build-from-source only, no Windows prebuilt path, and upgrades fail on a locked running binary). Note the draft is ~36 conflicting paths behind current main after v2.13.0 and this week's fleet — it needs a full rebase before review.

JoaoZaokk:
Yes — doing the full rebase. One decision I can't make unilaterally first, because it
changes how the rebase should land.

## The Windows CI job calls scripts `dev` no longer has

The trap is that this never shows up as a conflict. `.github/workflows/ci.yml` merges
cleanly, so the `windows` job survives the rebase intact and then fails at three of its
eight steps:

| step | script it calls | on `dev` |
|---|---|---|
| Controller checks and tests | `bun run test` | removed |
| Frontend and desktop checks and tests | `npm run test` | removed |
| Smoke packaged Windows desktop | `npm run desktop:smoke` | removed |

The other five pass. So I can hand you a rebased branch, but not one that can go green.
I'd rather agree the shape with you than guess and redo it.

## Why this is worth a minute even though it's my branch's problem

While rebasing I found that **no quality gate currently runs on Windows on `dev` at
all**. Three independent causes, each reproducible on a clean checkout:

1. `.githooks/*` are symlinks (mode `120000`). Git on Windows materialises them as
   22-byte text files, not executables, so any commit dies with
   `cannot spawn .githooks/pre-commit: No such file or directory`.
2. `scripts/project.mjs` is a symlink too — checks out as a 31-byte file, so
   `npm run check` and `npm run setup:git-hooks` cannot resolve it.
3. Even with a working shim, the hook dies at `spawnSync npm ENOENT` — Node on Windows
   will not spawn `npm.cmd` without `shell: true`.

This branch fixes all three. I mention it because it reframes the ask: the port isn't
requesting an exception to the testing policy, it's the thing that makes *any* gate —
test or not — executable on Windows in the first place.

And the fixes don't cost macOS anything. I ran the full gate set on a real M2
(macOS 26.5.2, node 25.9.0, bun 1.4.0): controller 129 tests, agent-runtime 99,
frontend 139, plus typecheck, lint and the four repo checks — all green. The executable
shims that replace the symlinks work natively there, and `npm run setup` completes
through them.

## Why the job leans on tests specifically

The three issues you consolidated are all behavioural and all invisible to typecheck and
lint:

- **#192** is a path-parsing contract. `path.basename` vs `split("/")` and
  `path.isAbsolute` vs `startsWith("/")` agree on POSIX and disagree on `C:\...`.
- **#210** is process lifecycle — whether a launched engine is still discoverable after
  a controller restart, and whether eviction kills the tree, is only observable by
  launching a real process and killing it.
- **#212** is installer behaviour under a locked running binary, a state that only
  exists at runtime on Windows.

## Options, in the order I'd pick them

1. **A Windows-only acceptance rail, named as such.** Not a unit suite: one job that
   launches a real engine, evicts it, and asserts the #192/#210/#212 behaviours. Never
   runs on macOS or Linux. My preference, and I'll keep it small.
2. **Keep the gates on the fork.** `dev` takes the source changes with no CI job; the
   fork catches regressions before they're proposed upstream.
3. **Land the port with no Windows gate at all.** Fastest to merge, and given #192,
   #210 and #212 are exactly the class of bug a static gate can't see, the one most
   likely to have them quietly return.

Tell me which and I'll shape the rebase to match. If it's (1), I'd also like to know
whether `NEVER WRITE TESTS` is meant to cover a real-process acceptance job or only unit
and snapshot suites — the wording covers both today and I don't want to argue the letter
of it.

## Separately

Two commits here were never Windows work, and I'm pulling them into their own PR rather
than smuggling them through: the configured remote-provider models change, and a
multi-GPU VRAM display fix mislabelled `fix(windows)` that a dual-GPU Linux host needs
identically. That also shrinks the rebase surface.

One of them is worth your attention independently of this PR: the shipped
`deepseek-v4-flash` starter preset on `dev`
(`controller/src/modules/studio/configs.ts:57`) points at
`http://pop-os-1.tailadb2c1.ts.net:8080/v1` — a personal Tailscale hostname, over plain
HTTP, that nobody outside that tailnet can reach. That's live on `dev` right now.

0xSero:
Deep review completed (all 110 files). No hostile content found — this is careful, well-guarded work: argument-array spawns throughout, the WSL wrapper passing data positionally, release downloads pinned to official ggml-org/sgl-project sources with validated tags, pinned CI toolchain, and no new dependencies. Thank you also for removing the leaked tailnet hostname from the DeepSeek preset (that fix has been landed separately via #440).

Before this can merge, three changes and a rebase:

1. `controller/src/modules/compute/engines/shared.ts` — `serveAddress` binds `0.0.0.0` for any non-`process` runtime. Under WSL *mirrored* networking that exposes the raw inference port (no auth of its own) to the LAN, bypassing the controller's API-key proxy. Bind the WSL NAT/localhost address, or gate mirrored mode explicitly.
2. `managed-llamacpp.ts` — please pin digests for the release archives; HTTPS+exact-name matching is good, checksums are better for binaries that get executed.
3. A "conservative" port shouldn't change non-Windows behavior: the instance-log path move, the model-visibility fallback, and the /v1/models provider-catalog append all take effect on macOS/Linux too (the last two are now superseded by main anyway). Please split anything intentional into its own PR.

Finally, main has moved substantially — the branch currently has ~52 conflicting paths including modify/delete conflicts, so it needs a full rebase onto current main before it's reviewable as a mergeable unit.

JoaoZaokk:
Closing this in favour of two small PRs against `main`, and answering the three asks from
your review. Thank you for reading all 110 files — the notes were specific enough to act
on, and two of the three turned out to be real defects.

## Why close rather than rebase

This targets `dev`, which has had no commits since 2026-08-19; everything from #427 on
targets `main`. A rebase is not the problem — the landing surface is. #443 deleted
`controller/src/modules/compute/launchers/process.ts`, `launchers/wsl2.ts`,
`core/process-platform.ts`, `compute/wsl-platform.ts` and `engines/llamacpp.ts`, dropped
`wsl2` from `RuntimeKind`, and reduced `EngineRuntimeKind` to a single member:

```ts
export type EngineRuntimeKind = "docker";
```

Most of what this PR carries is a second and third arm of a dimension that now has one.
That is a product decision I have no argument with, but it means the native-process, WSL2
and llama.cpp work here cannot be rebased onto `main` — there is nothing for it to attach
to. 184 paths conflict, and the conflicts are not textual.

## What replaces it

- **#450 — `fix(automation): make the repository automation runnable on Windows`.** Git
  on Windows defaults to `core.symlinks=false`, so a checkout writes `.githooks/*` and
  `scripts/project.mjs` out as text files containing their target paths. On a Windows
  clone of `main`, `node scripts/project.mjs doctor` is a `SyntaxError` and every hook,
  `npm run setup`, `npm run check` and `npm run build` fails with it. Shims instead of
  symlinks, plus three `win32`-only fixes for spawning `npm`/`npx`/`bun`, finding Python,
  and chmod. It also fixes `npm run build`, which dies in `bundleAgentRuntime` with
  `Agent runtime bundle failed with status unknown` — the status is `null` because the
  spawn never happened.
- **#451 — `feat(desktop): give the Windows package an icon and build it in CI`.**
  `electron-builder.yml` has declared `win: nsis x64` for a while; nothing built it and
  `win.icon` was never set, so the packaged app ships electron-builder's stock artwork.
  Icon plus a CI job shaped like the existing `desktop-package`.

Both are `main`-based, both are verified on Windows 11, and neither contains any of the
engine-runtime work.

## The three asks

**1. `serveAddress` binding `0.0.0.0`.** Confirmed, and you were right about the
consequence. Upstream's own expression was correct:

```ts
request.runtime === "docker" ? "0.0.0.0" : "127.0.0.1"
```

The port had inverted it to `request.runtime === "process" ? "127.0.0.1" : "0.0.0.0"` in a
one-line hunk inside an unrelated feature commit, with no rationale and no test, so every
WSL2 launch bound all interfaces. Under `networkingMode=mirrored` the distribution shares
the host's real NICs, which publishes the raw inference port on the LAN past the
API-key proxy. I measured it rather than argued it: on WSL 2.7.12 a listener bound to
`127.0.0.1` inside the distribution *is* reachable from Windows at `localhost`, so the
wildcard bought nothing and cost the isolation. Fixed with an exhaustive
`Record<EngineRuntimeKind, string>`, so a future runtime kind is a compile error rather
than a silent inheritance. Since this file no longer exists on `main`, the fix lives in my
edition — flagging it here only because the finding was yours.

**2. Pinned digests for the llama.cpp download.** Done, also edition-only now. The
installer previously ran `curl -L --fail` into staging, `tar -xf`, then executed
`llama-server.exe` with no integrity check of any kind. GitHub publishes a per-asset
`"digest": "sha256:…"` on the releases API, so every archive is hashed and compared before
extraction and the install fails closed both when the digest is absent and when it
mismatches. Worth knowing if the roster ever regrows a downloader: the same work turned up
that `/releases/latest` for `ggml-org/llama.cpp` now returns tag `v0.3.0`, whose entire
asset list is `["nightly-tag.txt"]` — the Windows binaries only exist on the rolling `b*`
builds, so any `latest`-based resolution silently finds nothing.

**3. Splitting the non-Windows behaviour changes.** Two of the three landed through #440,
with credit — thank you for that. The third has nothing left to split into. It kept live
compute-instance logs out of the retention sweep, and on `main` that defect cannot occur:
only `docker.ts` and `launcher.ts` remain, `launcher.ts` no longer has `readLogTail`,
`store.logPath` is declared and implemented with zero callers, and the docker launcher
reads through `docker logs` rather than a file. Nothing writes an instance log for the
sweep to unlink. So: no PR, rather than a PR for a defect that is not there.

## Where the rest goes

The native process launcher, the WSL2 launcher and llama.cpp continue in
[my edition](https://github.com/JoaoZaokk/local-studio-personal), which tracks the `dev`
lineage precisely because that is the last upstream state where those are first-class. No
hard feelings about the direction — docker-only is a coherent product, it just isn't the
one I need locally. I'll keep sending the portable pieces here.

PR #403, opened 2026-08-12
Session performance, litter-bridge removal, and the models page redesign

0xSero:
Three separate pieces of work, each landed as its own commit so they can be read — or reverted — independently.

## 1. Session performance (16 commits, `0960519d`…`402a1a94`)

Opening a session went from **~1.2s per process** — and **~32s** for a 3.56GB rollout — to **~30–60ms server-side, ~220ms cold in the browser**.

Six measured optimizations, each with a before/after number recorded in `docs/session-performance.md`:

| change | effect |
|---|---|
| fold canonical replay in place | 113ms → 78ms at 1600 messages |
| cache the active-branch walk per rollout | removes a full re-walk per open |
| resume the usage scan instead of restarting | O(file) → O(new bytes) |
| persist both caches across restarts | first open after a restart is warm |
| page the transcript from a de-noised sidecar | 145MB → 7.2MB read per session |
| scope the timeline merge cache to the transcript | stops eviction thrash on long sessions |

**Two correctness bugs surfaced along the way and are fixed here**, neither of which was a performance problem:

- Reloading a session silently truncated it, with no way to recover the rest (`496a7841`)
- The snapshot cache emptied itself whenever localStorage filled, wiping every other session to store one (`3f9877b7`)

**Four optimizations were rejected on evidence** and are documented so nobody rebuilds them: timeline virtualization (a cold open is 220ms and scrolling 6–9ms), a markdown fast path (≤0.69ms of a 3.5ms mount), skipping the branch walk for "linear" sessions (drops 69% of entries on a real rollout), and LRU for the merge cache (evicts exactly what it needs next).

## 2. Remove the litter-bridge gateway (`00f62e66`)

**−7,913 lines.** The bridge was a second, uncached session API — its own rollout discovery, header reads, pagination, cursors and event translation, sharing exactly one function with `sessions-store`.

It never carried load. The on-disk idempotency ledger holds three agent turns, all from 2026-07-20, and nothing since. Litter reaches the same sessions another way: its pi bridge reads `~/.pi/agent/sessions/` directly — 1,403 threads, 82 in the Local Studio workspace, 19 of those active *after* the gateway's last request. Watching the live daemon showed zero loopback connections; every socket went to a remote relay on :443.

`promptDurably` / `persistLitterPromptBoundary` go with it — they existed only so the bridge could correlate a mobile dispatch with a transcript entry, and nothing else read the marker they wrote. KittyLitter QR pairing is untouched; it shells out to the CLI and never involved this endpoint.

## 3. Models page on the Codex plugins pattern (`2b07d9fb`, and `09835a84` for the studio chrome)

Ported from the ChatGPT desktop bundle's plugins page: a section is a header plus a bounded card, rows live inside it, and the divider between rows is inset from the border rather than full-bleed.

Rows became flex instead of a two-column grid — the old grid pinned every label to a fixed 180px/260px track, which is what opened the dead gap mid-row. Expanded content now renders in a nested panel rather than hanging off a hardcoded left margin.

Our type ramp is deliberately kept rather than adopting Codex's absolute sizes; the structure was the problem, not the density.

Also fixes two overlays that used an opaque `--color-background` backdrop and made the app appear to vanish behind them.

## Verification

`npm run check` is green at every commit — typecheck, lint, cycles, structure gates, knip, jscpd, depcheck, controller standards, all package test suites, and a full production build.

Not visually verified: the models redesign passes typecheck, lint and build, but has not been looked at rendered.


0xSero:
Closing: main has moved to the consolidated tree landing in #412 (Models page rebuilt as one table, sidebar ordered by last prompt, dictation removed, packaging assertion corrected). Work from this branch that survived is already in that PR; anything still wanted should be reopened against the new main.

PR #409, opened 2026-08-16
Expose external provider models through GET /v1/models

### LINKED ISSUE #410: [Bug] External provider models are not surfaced in the model picker
## Use case

Controllers with enabled, API-key-configured external OpenAI-compatible providers currently expose those providers' models only through `GET /studio/provider-models`. The model picker discovers models through `GET /v1/models`, so models served by an external provider never appear in the picker even when the provider is configured and reachable.

Users who route chat/model selection through a remote controller with a connected external provider have no way to see or select those upstream models from the UI.

## Expected behavior

`GET /v1/models` should also return models from enabled, API-key-configured external providers so they show up in the model picker alongside the locally-discovered recipe/active-inference entries.

## Proposed solution

Append provider-discovered models to the `/v1/models` response using qualified IDs (`<provider-id>/<upstream-model-id>`), with per-provider discovery failures swallowed so one unreachable provider never fails the whole endpoint. The proxy already parses the qualified ID and rewrites the upstream request model, so no proxy changes are needed.

urucoder:
## Summary

Controllers with configured external OpenAI-compatible providers only surfaced those providers' models via `GET /studio/provider-models`. Since models are discovered through `GET /v1/models`, external provider models were invisible in the model picker even when the provider was configured and reachable.

## Changes

- **Shared discovery helper**: extracted `discoverProviderModels` and `enabledProvidersWithApiKey` into `controller/src/services/provider-routing.ts`, replacing the inline `providerModels` function previously duplicated in `provider-routes.ts`.
- **`/studio/provider-models`**: now consumes the shared helper; behavior unchanged.
- **`GET /v1/models`**: now also queries enabled, API-key-configured providers and appends their models as qualified IDs (`<provider-id>/<upstream-model-id>`), alongside existing recipe/active-inference entries. Per-provider discovery failures are swallowed (`Effect.option`) so one unreachable provider never fails the whole endpoint, matching `/studio/provider-models` semantics.
- No changes needed in the proxy (`openai-routes.ts`) — it already parses the qualified ID and rewrites the upstream request model correctly.

Example: with provider `my-provider` (base_url `http://<host>:8000`) enabled and reporting `example-model-1`, `GET /v1/models` now includes:

```json
{
  "id": "my-provider/example-model-1",
  "object": "model",
  "owned_by": "my-provider",
  "active": false,
  "metadata": { "vision": false }
}
```

## Validation

```bash
cd controller && bun run typecheck
cd controller && bun run lint
cd controller && bun run check
node scripts/project.mjs validate-contracts
node scripts/project.mjs validate-structure
node scripts/project.mjs audit-layout
```

Manually verified end-to-end against a live controller and a mock upstream provider server: `GET /v1/models` returned the qualified model ID, and a chat completion request using that ID was routed upstream with the model field rewritten back to the original provider model ID.

## UI changes

None — controller-only change. Effect is a discoverability fix for the existing model picker.

## Risks / rollout notes

- Purely additive to `/v1/models` response; existing consumers unaffected.
- Provider discovery adds one upstream HTTP call per enabled/configured provider per `/v1/models` request (10s timeout, best-effort, run concurrently) — negligible latency impact for typical provider counts.

Closes #410

### DIFF
--- a/controller/src/modules/models/routes.ts
+++ b/controller/src/modules/models/routes.ts
@@ -44,6 +44,7 @@ import { notFound } from "../../core/errors";
 import { findObservedInferenceProcess } from "../../core/function-observability";
 import { parseBooleanFlag } from "../../core/validation";
 import { fetchInference } from "../../http/local-fetch";
+import { discoverProviderModels, enabledProvidersWithApiKey } from "../../services/provider-routing";
 
 function isMockInferenceEnabled(): boolean {
   return parseBooleanFlag(process.env["LOCAL_STUDIO_MOCK_INFERENCE"]);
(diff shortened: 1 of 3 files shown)


urucoder:
Closes #410

0xSero:
Re-implemented in #428 together with #380's best parts: discovery lives in provider-routing (this PR's placement), /v1/models appends providerId/model entries with active:false so a provider model can't steal the default from the running local model, per-provider 3s deadlines and a 60s/15s cache keep dead providers from stalling the picker. Fixes #410.

PR #416, opened 2026-08-19
fix(usage): restore the profile header and stat grid above the tabs

0xSero:
The overnight rebuild kept the data and threw away the page: the profile avatar and its editable name, the headline token count, and the six-cell stat grid were all replaced by a generic tabbed header — and the per-tab strip then repeated four of the six numbers it had just removed.

Header and grid are back above the tab strip, where they belong (those six are true of the whole retention window, not of any one tab). The tabs stay. The Models strip is trimmed to what only that tab can say: token mix, cache hit rate, decode.

The grid uses `gap-px` over a border-coloured ground instead of `divide-x` — a divided grid draws a stray left edge on the first cell of every wrapped row, so it only looked right at the single breakpoint where all six cells fit on one line. Verified at 6-up and wrapped 3×2. Skeleton reshaped to match, so nothing moves when data lands.

### DIFF
--- a/frontend/src/features/usage/usage-activity-tab.tsx
+++ b/frontend/src/features/usage/usage-activity-tab.tsx
@@ -15,7 +15,6 @@ import {
   TableNotice,
   type Stat,
 } from "@/features/recipes/recipes-content/catalog-table-shell";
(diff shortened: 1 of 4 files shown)


0xSero:

PR #417, opened 2026-08-19
perf: stop one dead controller stalling the model list, ship catalog logos

0xSero:
Both measured, not guessed.

## Composer model list: 10.5s → capped at 4s

`/api/agent/models` took **10.505s, 10.507s, 10.509s** — that consistency is a timeout being waited out, not slow work. The configured controller answers `/v1/models` in **20ms**, so almost none of it was real.

`fetchModelsFromController` used a bare `fetch` with no deadline, and `Promise.allSettled` waits for every controller, so the picker cannot show one model until the slowest resolves. Of the four saved controllers here:

| pop-os-1 | omarchy | spark-de5c | spark-2822 |
|---|---|---|---|
| 0.30s | 0.02s | **never answers** | 1.23s |

Each controller now gets 4s; a slower one is reported failed and the rest still load. The user's config is untouched — the mechanism was the bug.

## Models page icons

Publisher avatars resolved through Hugging Face at runtime. The request could not even *start* until the catalog loaded and the table rendered (492ms), then cost 160–470ms per owner — so a cold load showed coloured initials flipping to logos.

The curated catalog has a fixed set of publishers, so their logos now ship as static assets (15 files, 92 KB). Measured on a production build with a cold cache:

| | before | after |
|---|---|---|
| Hugging Face round trips | 8 | **0** |
| slowest icon | 466ms | **41ms** |
| all icons settled | ~958ms | **598ms** |

Unknown owners from Hugging Face search still use the cached proxy — right tool for the long tail.


0xSero:

PR #418, opened 2026-08-19
fix(installer): delimit port expansion before ellipsis

Morrowmake:
## Summary

Delimit the controller port expansion before the adjacent Unicode ellipsis. Under Bash strict unset-variable handling, the undelimited form is parsed as a different variable name, causing the installer to exit before its health check and registration marker.

## Validation

```bash
bash -n scripts/install-controller.sh
bash -uc 'PORT=8080; printf "%s\n" "waiting for controller on :${PORT}…"'
npm run check
```

## UI changes

None.

## Risks / rollout notes

Minimal shell-only change. Controller startup, health-check behavior, and configuration remain unchanged.

### DIFF
--- a/scripts/install-controller.sh
+++ b/scripts/install-controller.sh
@@ -204,7 +204,7 @@ else
 fi
 
 # --- health ------------------------------------------------------------------
-log "waiting for controller on :$PORT…"
+log "waiting for controller on :${PORT}…"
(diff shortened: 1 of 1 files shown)


0xSero:
Cherry-picked into #428 (the single real commit; the branch's other 25 commits were a stale base). Thanks — the fix was exactly right.

PR #420, opened 2026-08-19
fix: quiet the bell, stop composer hydration churn, surface automation threads

0xSero:
Four fixes across the shell, composer and automations.

## Nav header
- **The bell looked permanently hovered** — it carried `aria-pressed:bg-(--link)/15`, so it sat filled blue whenever the notifications view was open. Pressed is now a foreground shift only, matching the buttons beside it. Verified: background computes to `rgba(0,0,0,0)` at rest with the view open.
- **The blue dot is gone**, replaced by actual session state: a pulsing count of live runs, or a quieter count of runs that finished while you were away. Nothing renders when nothing is happening, so the resting nav is quieter than before.

## Composer hydration
The project label is built from two client-only sources — the pane's restored view state and the projects store, which seeds from `localStorage` **synchronously at creation**. The server can know neither, so it rendered "Choose project" while the client's first render said the project name. React answers a hydration mismatch by discarding and re-rendering the subtree, and that subtree is the composer, on every single load.

Went from throwing on every page load to **zero hydration errors**, confirmed in the browser.

## Automations
- **Threads were hidden.** Run history linked to a transcript only when a run recorded *both* a session id and a project id. The scheduler derives `projectId` by matching the run's cwd against registered projects, so an automation pointed anywhere else recorded `null` and the UI claimed "Transcript unavailable" for a thread that existed and was perfectly reachable. The session id alone opens it — `workspaceNavigation` already treats the project as optional.
- **Runs no longer die on an evicted model.** A schedule outlives the model it was written against. If the configured model is not live, the run now uses whatever is and logs the substitution; if nothing is live, the configured id is kept so the failure still names the model you picked.

### DIFF
--- a/frontend/src/features/agent/automations/automation-editor.tsx
+++ b/frontend/src/features/agent/automations/automation-editor.tsx
@@ -314,10 +314,14 @@ function RunHistory({ automation }: { automation: Automation }) {
       </div>
       <div className="divide-y divide-(--ui-separator) border-y border-(--ui-separator)">
         {automation.runs.map((run, index) => {
-          const transcriptHref =
-            run.piSessionId && run.projectId
-              ? `/agent?project=${encodeURIComponent(run.projectId)}&session=${encodeURIComponent(run.piSessionId)}&replace=1`
-              : null;
+          // The session id is what opens the thread; the project only preselects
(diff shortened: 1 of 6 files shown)


0xSero:

PR #435, opened 2026-08-22
fix(providers): expose configured remote models and report the full VRAM pool

JoaoZaokk:
Two commits that were sitting on the Windows port branch (#395) but were never Windows
work. Pulling them out rather than smuggling them through a platform PR. Nothing here is
platform-specific.

## `fix(providers): expose configured remote models`

`GET /v1/models` listed only local recipes, so a user whose providers are all remote saw
an empty catalogue and an empty model picker. This merges the configured providers' own
`/v1/models` into the listing as `provider/modelId`, marked remote, and lets the picker
fall back to those when the controller has none of its own.

The provider fetch moves out of the studio provider route into a shared service so the
two callers stop diverging, and both now route through one helper that appends the
OpenAI-compatible `/v1` prefix. That helper being the only place a provider `base_url`
gets a suffix is what makes the next part safe.

**Worth a look independently of the rest:** the shipped `deepseek-v4-flash` starter
preset in `controller/src/modules/studio/configs.ts` points at
`http://pop-os-1.tailadb2c1.ts.net:8080/v1` — a personal Tailscale hostname, over plain
HTTP, that nobody outside that tailnet can reach. That is live on `dev` right now. This
repoints it at `https://api.deepseek.com`. Installs that already stored the old value get
repointed when the preset is reconnected.

## `fix(setup): report the full multi-GPU VRAM pool`

The hardware summary read `memory_total_mb` off the first GPU only, so a two-card machine
advertised half its pool during setup and the model recommendation was sized against the
wrong number. This sums every reported GPU. Apple Silicon keeps reading the single
unified entry.

The controller already returned every GPU, so this is display arithmetic, not telemetry —
a dual-GPU Linux or Windows host is affected identically. It landed on the port branch
under a `fix(windows)` scope that was simply wrong.

## Notes

- Branched from `dev`, 9 files, no new dependencies.
- The test files that accompanied both commits on the port branch are **omitted** here,
  since `dev` has no test rail. They remain on the port branch if you ever want them.
- Verified against a pristine `origin/dev` baseline: `tsc --noEmit` output for the
  frontend is byte-identical to the baseline, and the controller typechecks at exit 0.
  The 12 pre-existing errors under `shared/` are unchanged and untouched by these files.
- Removing these two commits also shrinks the rebase surface of #395.

### DIFF
--- a/controller/src/modules/models/routes.ts
+++ b/controller/src/modules/models/routes.ts
@@ -43,6 +43,7 @@ import { isRecipeRunning } from "./recipes/recipe-matching";
 import { notFound } from "../../core/errors";
 import { findObservedInferenceProcess } from "../../core/function-observability";
 import { fetchInference } from "../../http/local-fetch";
+import { fetchConfiguredProviderModels } from "../../services/provider-models";
 
 function isRecord(value: unknown): value is Record<string, unknown> {
   return Boolean(value) && typeof value === "object" && !Array.isArray(value);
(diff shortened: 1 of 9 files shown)


0xSero:
Deep review completed — nothing hostile found, and thank you for catching the leaked tailnet hostname in the DeepSeek preset. The catalog-merge plumbing here was superseded by the cached provider-catalog merge that landed on main (which also marks remote models inactive rather than active, and avoids the uncached fan-out on the model-picker hot path), so rather than ask for a rebase, the four still-relevant pieces were ported onto current main in #440 with authorship credit: the preset URL fix, /v1-tolerant base URLs, the full-VRAM-pool sum, and the empty-picker fallback. This PR will close when #440 merges.

PR #439, opened 2026-08-24
fix(agent): require strict goal outcome sentinels

0xSero:
Lands #430 by @moortekweb-art with a regenerated agent-runtime lockfile and one hardening tweak on top: the blocked sentinel now requires a non-identifier boundary, so a final line like GOAL_BLOCKED_BY_X no longer parses as blocked with a garbled reason.

Sentinels must now sit alone on the reply's final line, which shrinks the surface where quoted tool output or page text could falsely settle a goal. All 19 goal tests pass; the pinned typescript 5.9.3 lockfile entry was verified against the npm registry's published integrity hash.

Closes #430.

### DIFF
--- a/services/agent-runtime/bun.lock
+++ b/services/agent-runtime/bun.lock
@@ -19,6 +19,7 @@
       "devDependencies": {
         "@types/node": "24.6.0",
         "@types/proper-lockfile": "4.1.4",
+        "typescript": "5.9.3",
       },
     },
   },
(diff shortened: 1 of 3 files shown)


moortekweb-art:
Thanks for landing this. I independently validated the merged parser on commit 6a6ece0 (the focused goal-protocol suite passes: 11/11), and found one remaining boundary case in the same class as GOAL_BLOCKED_BY_X:

```text
goalOutcomeFromText("GOAL_BLOCKEDé")
=> { kind: "blocked", reason: "é" }
```

GOAL_BLOCKED_BY_X correctly returns null. The current negative lookahead is ASCII-only, while the optional separator still permits an attached suffix. The robust grammar fix is to require either end-of-line or an explicit separator (space/tab, colon, hyphen, en dash, or em dash) before a blocked reason, and keep BLOCKED_SENTINEL_LINE_RE and SENTINEL_LINE_RE in lockstep. That rejects Unicode letters, combining marks, ZWJ/ZWNJ, dollar-sign suffixes, emoji, and attached punctuation by construction while preserving GOAL_BLOCKED, GOAL_BLOCKED reason, GOAL_BLOCKED:reason, and GOAL_BLOCKED — reason.

I did not open a follow-up PR because AGENTS.md requires branching from origin/dev, and shared/agent/goal-protocol.ts is not currently present on origin/dev. Happy to prepare the narrow patch when that branch contains the merged protocol file.

moortekweb-art:
@0xSero Michael has approved preparing the narrow follow-up. Since AGENTS.md requires contributors to branch from and target dev, but shared/agent/goal-protocol.ts currently exists only on main, would you prefer that I treat this as a main-targeted release hotfix, or wait until the protocol is synced to dev? I will keep the change limited to the blocked-sentinel separator grammar and direct runtime verification.

PR #443, opened 2026-08-25
refactor: registry-backed recipes, docker-only engine roster, unified passthrough proxy

0xSero:
The omarchy program, complete. Five commits, each landed with all gates green and live-boot verification:

1. **Recipes live in the model registry** — `data_dir/model-index.json` carries launchable entries beside catalog tiers; SQLite `recipes` becomes a one-shot, stamped migration source and rollback path. RecipeStore keeps its Effect interface so nothing above it changed.
2. **The bridge is dead** — compute owns the one-active-model surface (`compute.model`); the four legacy lifecycle routes moved into compute with identical wire shapes; the crash-loop budget moved with them.
3. **One passthrough proxy** — /v1/chat/completions, /v1/responses, and /v1/messages are served by a single pipeline: recipe/provider routing, the running-model gate, byte-for-byte forwarding, and usage recording per dialect (verified against a mock engine: six paths byte-identical, six records with exact token mappings). The 1,650-line compensation layer for parserless engines is deleted.
4. **Docker-only engines, roster of three** — vllm + sglang + exllamav3 (newly launchable via TabbyAPI); llamacpp and mlx leave the product. The runtime existence matrix (venvs, binaries, discovery, probes, install locks, ~3,700 lines) is replaced by "pinned image, pulled" (~280 lines). Legacy recipes with dropped backends are preserved in the registry and reported at boot, never destroyed.
5. **Frontend follow-through** — roster editor, image rows with Pull actions, docker default runtime for new recipes, GGUF/MLX flows removed, three-backend status maps.

Net: **−7,164 lines** across the branch. Deployment note for pop-os: evict the running model before restarting the controller on this code; the next launch requires docker + nvidia-container-toolkit, and live recipes migrate into the registry on first boot (recipe DB backup protocol applies).


Raynos:
This pull requests deletes the MLX support I need to run Local Studio on my macbook pro.

PR #445, opened 2026-08-27
Fix: Start the controller during desktop development

gildrb:
desktop:dev launched Electron and Next.js, but without the controller, which repeatedly caused proxy 500 errors in the console. This change starts the controller with the desktop development stack to quiet down the logs for desktop contributers.


gildrb:
Replaced by #446. This PR was opened from fork `main` to upstream `main` after a `dev`-based branch was merged into the fork, which introduced unrelated changes and a malformed generated automation file. #446 uses the isolated feature branch and targets `dev`.

PR #446, opened 2026-08-27
fix: start the controller during desktop development

gildrb:
## Problem

`desktop:dev` opened Electron and started the Next.js frontend without starting the controller. On a fresh contributor setup, the renderer immediately requested controller endpoints such as `/status`, `/recipes`, and `/events`. Those requests failed and continuously filled the development log with proxy 500 errors.

The documented root controller commands also used a Bun argument order that Bun 1.3.14 rejects, so trying to start the missing controller printed Bun help instead.

## Change

- start the controller as part of `desktop:dev`
- enable desktop proxy behavior for the Next.js development process
- use Bun 1.3.14-compatible `--cwd` argument ordering in the root controller scripts

After this change, contributors can launch the complete live desktop stack with one command and get a clean, useful development log instead of connection-error spam.

## Validation

- controller TypeScript check
- Electron main/preload build
- frontend static and cleanup gates
- production frontend build and standalone assertion


gildrb:
Opened against upstream by mistake. This change belongs in the fork and has already been merged there.

PR #453, opened 2026-08-30
fix(api): keep the home directory out of what nft has to trace

### LINKED ISSUE #452: next build fails on windows-latest: EPERM globbing the runner home directory
`next build` does not complete on a `windows-latest` GitHub runner. It fails during
compilation with an `EPERM` while globbing the runner user's home directory — a directory
the build has no reason to be in, on a different drive from the workspace.

## What happens

Workspace is `D:\a\local-studio\local-studio`. `npm run setup` succeeds. Then:

```
> node ../scripts/project.mjs prepare-next && npm --prefix ../services/agent-runtime run bundle && next build --webpack && ...

Bundled 2404 modules in 574ms
  standalone.mjs  7.25 MB  (entry point)
Packaged portable browser runtime: playwright-core, chromium-bidi, mitt, devtools-protocol, ...
▲ Next.js 16.2.12 (webpack)
  Creating an optimized production build ...
glob error [Error: EPERM: operation not permitted, scandir 'C:\Users\runneradmin\Application Data'] {
  errno: -4048,
  code: 'EPERM',
  syscall: 'scandir',
  path: 'C:\Users\runneradmin\Application Data'
}
Failed to compile.

Error: EPERM: operation not permitted, scandir 'C:\Users\runneradmin\Application Data'

> Build failed because of webpack errors
```

`C:\Users\<user>\Application Data` is the legacy junction Windows keeps for pre-Vista
compatibility. It points at `AppData\Roaming`, which contains it, so Windows denies
enumeration deliberately to stop infinite recursion. Any recursive walk that reaches a
Windows home directory hits it.

Full run: <https://github.com/sybil-solutions/local-studio/actions/runs/33286898609>

## What I could rule out

- **Not the junction being unreadable per se.** `readdirSync` on `%USERPROFILE%\Application
  Data` throws the same `EPERM` on my own Windows 11 machine, where the build succeeds.
  Something globs the home directory on the runner and does not locally.
- **Not the `@oxc-resolver` install warning** earlier in the same job
  (`npm warn cleanup Failed to remove some directories … EPERM … binding-wasm32-wasi`).
  `oxc-resolver` is reached only through `knip`, a devDependency, and is not part of the
  build graph.
- **Not the agent-runtime bundle.** It completes and prints its runtime list before Next
  starts.
- **Not this repository's Windows packaging.** `electron-builder --dir --win --x64`
  succeeds locally once the build has produced the standalone output, `afterPack`
  assertion and 5 MB asar ceiling included.

The `glob error` prefix and the timing (after `Creating an optimized production build`)
point at Next's own globbing of `outputFileTracingIncludes` / `outputFileTracingExcludes`
against an `outputFileTracingRoot` that resolves somewhere it should not on the runner.
`next.config.ts` sets both, with patterns like `"../controller/**/*"` and `"../*.md"`,
and does not set `outputFileTracingRoot` explicitly. I could not confirm that without
iterating on a runner, and did not want to guess at a config change that would also change
tracing on macOS and Linux.

## Reproducing

Adding this job is enough:

```yaml
  desktop-package-windows:
    if: github.event_name != 'schedule'
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
        with:
          node-version: 22.19.0
      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
        with:
          bun-version: 1.3.14
      - name: Install repository dependencies
        run: npm run setup
      - name: Build unsigned package
        env:
          CSC_IDENTITY_AUTO_DISCOVERY: "false"
        run: npm --prefix frontend run desktop:pack
```

It needs #450 first — without it `npm run setup` and `npm run build` fail earlier on
Windows for unrelated, fixable reasons (symlinked entry points, and `spawnSync("bun", …)`
in `bundleAgentRuntime`).

## Why I am reporting rather than fixing

I hit this adding that job to #451 and withdrew the job rather than merge a red check for
a problem the PR did not cause. It does not reproduce on my Windows 11 machine with the
same Node 22, npm 10.9.7 and Bun 1.3.14, so I have no way to test a fix short of pushing
candidates at CI. Happy to do that if it is wanted — just did not want to guess at
`outputFileTracingRoot` in a PR about an icon.

JoaoZaokk:
Fixes #452.

`next build` fails on Windows when the workspace lives outside the user profile:

```
glob error [Error: EPERM: operation not permitted, scandir 'C:\Users\<user>\Application Data']
Failed to compile.
```

## Cause

Next traces the route handlers with `@vercel/nft`, which constant-folds `os.homedir()` into a literal path. It cannot resolve what `readdir` and `statSync` are finally given, so it widens the call to `<home>/**/*` and hands that to glob. A Windows user profile contains `Application Data` and `Cookies`, pre-Vista junctions the OS denies enumeration on, and glob v7 propagates `EPERM` for anything that is not `ENOENT` / `ELOOP` / `ENAMETOOLONG` / `UNKNOWN`.

Two routes reach it: `api/agent/directories`, where the default root and the fallback root both fold, and `api/agent/terminal/resolve-cwd`, through tilde expansion and the empty target. Either one alone is enough, because the glob fires per route.

This is also why `outputFileTracingExcludes` does not help: the glob throws during the walk, before there is any result left to filter.

## Fix

Read the variable Node itself reads. `os.homedir()` returns `$HOME` on POSIX and `%USERPROFILE%` on Windows before falling back to a system lookup, so reading that variable first gives the same value at runtime while staying opaque to the tracer. `os.homedir()` remains the last resort, and the tracer folds it harmlessly because a union with an unknown is unknown.

Each spelling was traced directly through the vendored `@vercel/nft` before this was written: `os.homedir()` reaching either root emits the glob, the env-first form emits nothing, and `os.homedir()` elsewhere in the same module is harmless.

## Verified

Windows 11, workspace on `F:`, profile on `C:`:

| | result |
| --- | --- |
| before | `EPERM: scandir 'C:\Users\<user>\Application Data'`, build fails |
| after | build completes |

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,1 +1,5 @@
-../scripts/project.mjs
+#!/usr/bin/env node
+import("../scripts/project.mjs").catch((error) => {
+  console.error(error);
+  process.exit(1);
+});
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,1 +1,5 @@
-../scripts/project.mjs
+#!/usr/bin/env node
+import("../scripts/project.mjs").catch((error) => {
+  console.error(error);
+  process.exit(1);
+});
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,1 +1,5 @@
-../scripts/project.mjs
+#!/usr/bin/env node
+import("../scripts/project.mjs").catch((error) => {
+  console.error(error);
+  process.exit(1);
+});
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,26 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+
+  windows-automation:
+    name: Repository automation (Windows)
+    if: github.event_name != 'schedule'
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 22.19.0
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: 1.3.14
+      - name: Toolchain check
+        run: npm run doctor
+      - name: Git hook registration
+        run: node scripts/project.mjs setup-hooks
+      - name: Automation layout gate
+        run: node scripts/project.mjs audit-layout
+
   trufflehog:
     name: Secret Scanning (TruffleHog)
     runs-on: ubuntu-latest
--- a/frontend/desktop/automation/agent-runtime.mjs
+++ b/frontend/desktop/automation/agent-runtime.mjs
@@ -14,7 +14,7 @@ import {
   writeFileSync,
 } from "node:fs";
 import path from "node:path";
-import { repoRoot } from "./lib.mjs";
+import { platformInvocation, repoRoot } from "./lib.mjs";
 
 const packageDir = path.join(repoRoot, "services", "agent-runtime");
 const distDir = path.join(packageDir, "dist");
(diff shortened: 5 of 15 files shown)


JoaoZaokk:
Rebased onto #450. Only three files here are mine — `frontend/src/app/api/_lib/home-root.ts` and the two route handlers; the rest of the diff is #450, which this stacks on because a Windows build is impossible on plain `main`: the `scripts/project.mjs` symlink checks out as text, its relative imports then resolve from the wrong directory, and the agent-runtime bundle dies on `spawnSync("bun", …)`. All three are what #450 fixes.

